PR #8, opened 2023-10-08
pom.xml fixed according to 21 Java

Y-Pazniak:
packages was extracted from the root
big Storage enum was divided into the small parts
constants was renamed into better (in my opinion) way into enums constans was divided into groups
main was extracted from the controller package and moved to the root createSummaryString was renamed


src/main/java/by/custom/utilcalculator/service/Bot.java
```diff
@@ -196,23 +201,23 @@ private String getStringOptionsForEngineVolume() {
     private String getStringOptionsForAgeAuto() {
         return "." +
                 "\n" +
-                Storage.AGE_OF_AUTO;
+                Questions.AGE_OF_AUTO;
     }
 
     private String getStringOptionsForPrice() {
```
MaximPilipeiko, line 207: I proceed to see `String` in many methods. Is it really needed?

src/main/java/by/custom/utilcalculator/service/CalculatorPassenger.java
```diff
@@ -26,10 +28,11 @@ public String calculate(CountryOrigin countryOrigin, OwnersType ownersType, Type
             case null -> {
                 return "calculator: countryOrigin is null";
             }
-            case default -> {
+            default -> {
                 return "calculator: calculate() unknown error";
             }
         }
+//        return "calculator: calculate() unknown error";
```
MaximPilipeiko, line 35: why not remove it?

MaximPilipeiko (approved):

src/main/java/by/custom/utilcalculator/service/Bot.java
Y-Pazniak: We don't need it, sorry, my fault - didn't clean it yestarday. Today's fix contains less ...String... methods names. I still have two methods with this words, but I will try to fix it too.

src/main/java/by/custom/utilcalculator/service/CalculatorPassenger.java
Y-Pazniak: I love this messages a lot. They really helped me when I changed Java-version, and it was hard for me to remove it. Anyway, new commit doesn't contain this comments. Sorry for this mess. I will try to put my code in better order.

PR #10, opened 2023-10-13
Calculation develop bot segregation

Y-Pazniak:
Bot class was segregated.
Added classes for creating languages bundle, for checking input messages and for creating answers.
Unused files with constants were removed


src/main/java/by/custom/utilcalculator/service/MessagesChecker.java
```diff
@@ -0,0 +1,50 @@
+package by.custom.utilcalculator.service;
+
+import by.custom.utilcalculator.controller.Bot;
+import by.custom.utilcalculator.repository.resources.Commands;
+import org.telegram.telegrambots.meta.api.methods.send.SendMessage;
+import org.telegram.telegrambots.meta.api.objects.Message;
+import org.telegram.telegrambots.meta.api.objects.Update;
+
+public class MessagesChecker {
```
MaximPilipeiko, line 9: I like this solution. Maybe better to rename it to `MessageContoller` and move it to the controller package.

src/main/java/by/custom/utilcalculator/service/MessagesChecker.java
```diff
@@ -0,0 +1,50 @@
...
+            return messagesCreator.getSorryMessage(message);
+        }
+    }
+
+    public SendMessage getCheckInputMessageAndGetAnswer(Message message) {
+        String usersMessage = message.getText();
+        SendMessage sendMessage;
+        switch (usersMessage) {
+            case Commands.START -> sendMessage = messagesCreator.getGreetingMessage(message);
+            case Commands.EAES, Commands.OTHER_COUNTRIES -> sendMessage = Bot.getBot().getOriginCarMessage(message);
```
MaximPilipeiko, line 33: It's not correct to call Bot methods from this class. Usually, lower-level classes don't know about higher. 
We will discuss it on call.

src/main/java/by/custom/utilcalculator/controller/Bot.java
```diff
@@ -0,0 +1,139 @@
...
+import by.custom.utilcalculator.service.MessagesChecker;
+import by.custom.utilcalculator.service.MessagesCreator;
+import org.telegram.telegrambots.bots.TelegramLongPollingBot;
+import org.telegram.telegrambots.meta.api.methods.send.SendMessage;
+import org.telegram.telegrambots.meta.api.objects.Message;
+import org.telegram.telegrambots.meta.api.objects.Update;
+import org.telegram.telegrambots.meta.exceptions.TelegramApiException;
+
+public class Bot extends TelegramLongPollingBot {
+    private CountryOrigin countryOrigin = null;
```
MaximPilipeiko, line 14: Please, test if two users simultaneously use the Bot. I think the conflict is possible.

MaximPilipeiko (changes requested):

src/main/java/by/custom/utilcalculator/service/MessagesChecker.java
Y-Pazniak: It was renamed to MessageRouter

src/main/java/by/custom/utilcalculator/service/MessagesChecker.java
Y-Pazniak: I agree. Next pull request is about fixing this issue. I hope, I did everything OK

src/main/java/by/custom/utilcalculator/controller/Bot.java
Y-Pazniak: I didn't test it. But i am sure this problem will be, because we have only one singleton with fields. I will fix it soon (i hope)

PR #11, opened 2023-10-18
Calculation develop bot segregation refactoring v1

Y-Pazniak:
Fixed structure fail with wrong relationships between higher and lower classes. 
Fixed naming in laguage resource.


src/main/java/by/custom/utilcalculator/Main.java
```diff
@@ -9,12 +10,22 @@
...
-            TelegramBotsApi bot = new TelegramBotsApi(DefaultBotSession.class);
-            bot.registerBot(Bot.getBot());
+            TelegramBotsApi telegramBot = new TelegramBotsApi(DefaultBotSession.class);
+            telegramBot.registerBot(localBot);
         } catch (TelegramApiException e) {
             e.printStackTrace();
         }
     }
+
+    private static void singletonInits() {
```
MaximPilipeiko: I'm not sure that it makes sense.
We will init classes in Main but in a different way.

src/main/java/by/custom/utilcalculator/service/MessageRouter.java
```diff
@@ -1,22 +1,25 @@
 package by.custom.utilcalculator.service;
```
MaximPilipeiko: Please, move it under `controller` package
```suggestion
package by.custom.utilcalculator.controller;
```

src/main/java/by/custom/utilcalculator/service/BotFieldsManager.java
```diff
@@ -0,0 +1,100 @@
...
+    public BotFieldsManager() {
+        botEntity = BotEntity.getInstance();
+        messagesManager = MessagesManager.getInstance();
+    }
+
+    public static BotFieldsManager getInstance() {
+        return BotFieldsManagerHolder.BOT_FIELDS_MANAGER;
+    }
+
+    public SendMessage setCarOriginFieldAndGetNextMessageForUser(Message inputMessage) {
```
MaximPilipeiko: I prefer that BotFieldsManager doesn't know about telegram-specific classes, `SendMessage`, `Message`.
```suggestion
    public String setCarOriginFieldAndGetNextMessageForUser(User sender, command: Commands) {
```

src/main/java/by/custom/utilcalculator/directory/resources/Commands.java
MaximPilipeiko, line 1: It's more common practice to use singular name for enum

src/main/java/by/custom/utilcalculator/directory/resources/Prices.java
```diff
@@ -1,4 +1,4 @@
-package by.custom.utilcalculator.repository.resources;
+package by.custom.utilcalculator.directory.resources;
 
 public class Prices {
```
MaximPilipeiko: It's more common practice to use singular name for enum 
```suggestion
public class Price {
```

src/main/java/by/custom/utilcalculator/service/BotFieldsManager.java
MaximPilipeiko: I think in this case we can use a simpler but enough informative method name. The same for other methods.
```suggestion
    public String processCarOrigin(String inputMessage) {
```

src/main/java/by/custom/utilcalculator/directory/BotEntity.java
```diff
@@ -0,0 +1,63 @@
+package by.custom.utilcalculator.directory;
+
+import by.custom.utilcalculator.directory.steps.*;
+
+public class BotEntity {
```
MaximPilipeiko: Maybe `UserProgress` will be more informative.

src/main/java/by/custom/utilcalculator/service/BotFieldsManager.java
```diff
@@ -0,0 +1,100 @@
...
+                botEntity.getCountryOrigin(), botEntity.getOwnersType(), botEntity.getTypeOfEngine(), botEntity.getVolumeOfEngine(), botEntity.getCarAge());
+    }
+
+    public SendMessage setTypeOfOwnerFieldAndGetNextMessageForUser(Message inputMessage) {
+        cleanStepsAfterCurrent(2);
+        switch (inputMessage.getText()) {
+            case Commands.JURIDICAL_PERSON -> botEntity.setOwnersType(OwnersType.JURIDICAL);
+            case Commands.PHYSICAL_PERSON -> botEntity.setOwnersType(OwnersType.PHYSICAL);
+        }
+        return messagesManager.createSendMessage(inputMessage.getChatId().toString(),
```
MaximPilipeiko: Please, move `createSendMessage(` to MessageRouter

src/main/java/by/custom/utilcalculator/service/MessagesManager.java
```diff
@@ -1,131 +1,133 @@
...
-import by.custom.utilcalculator.repository.resources.Commands;
-import by.custom.utilcalculator.repository.steps.*;
+import by.custom.utilcalculator.directory.BotEntity;
+import by.custom.utilcalculator.directory.resources.Commands;
+import by.custom.utilcalculator.directory.steps.*;
 import org.telegram.telegrambots.meta.api.methods.send.SendMessage;
 import org.telegram.telegrambots.meta.api.objects.Message;
 
-public class MessagesCreator {
+public class MessagesManager {
```
MaximPilipeiko: Now I think we have a lot of managers and `MessageCreator` reflects a sense better.

MaximPilipeiko (changes requested):

src/main/java/by/custom/utilcalculator/Main.java
Y-Pazniak: Removed it. Waiting for instructions to fix it. 
To be honest, when I did this, I clearly understood that I was doing useless thing. But I had to do something to try init Singletons in the main class, but I hadn't another ideas how to handle this request except this silly way.

src/main/java/by/custom/utilcalculator/service/MessageRouter.java
Y-Pazniak: Done.

src/main/java/by/custom/utilcalculator/service/BotFieldsManager.java
Y-Pazniak: I fixed it, but in a little another way. Now BotFieldsManager doesn't know anything about telegram-specific classes. 
Your suggestion is: 
public String setCarOriginFieldAndGetNextMessageForUser**(User sender, command: Commands)** {
My variant is: 
public String setCarOriginFieldAndGetNextMessageForUser(**String command**) {
I am not sure if  I am right. But in my opinion we don't need to add here anything about sender - we can add it in the MessageRouter class. And we don't need to give array: we send only one command. In my opinion my variant is better, because we send less options into the methods. But probably I made a mistake. Please, check it.

src/main/java/by/custom/utilcalculator/directory/resources/Commands.java
Y-Pazniak: Fixed.

src/main/java/by/custom/utilcalculator/directory/resources/Prices.java
Y-Pazniak: Fixed.

src/main/java/by/custom/utilcalculator/service/BotFieldsManager.java
Y-Pazniak: Fixed. 
There are two operations in this methods. First operation - to change the field of Bot object. 
`case Command.EAES -> botEntity.setCountryOrigin(CountryOrigin.EAES);`
The second operation is to return String with message to send user like an answer.
`return messagesManager.createSendMessage();`
Is it OK or we must to split this method in two different methods?

src/main/java/by/custom/utilcalculator/directory/BotEntity.java
Y-Pazniak: Done.

src/main/java/by/custom/utilcalculator/service/BotFieldsManager.java
Y-Pazniak: I think, we don't need this method anymore. I have removed it at all. Now we can call MessagesManagers methods from BotFieldsManager class. We don't need this wrapper createSendMessage() anymore, I guess. All it's logic is in the MessageRouter router and MessageManager. So I removed it.

src/main/java/by/custom/utilcalculator/service/MessagesManager.java
Y-Pazniak: Renamed it to MessagesCreator

src/main/java/by/custom/utilcalculator/controller/MessageRouter.java
```diff
@@ -0,0 +1,72 @@
...
+        botFieldsManager = BotFieldsManager.getInstance();
+    }
+
+    public static MessageRouter getInstance() {
+        return MessagesCheckerHolder.MESSAGES_CHECKER;
+    }
+
+    public SendMessage checkMessageBeforeExecutionAndGetResult(Update update) {
+        Message message = update.getMessage();
+        if (message.hasText()) {
```
MaximPilipeiko, line 25: Just it looks nicer. 
```suggestion
        if (!message.hasText()) {
            SendMessage sorrySendMessage = new SendMessage();
            sorrySendMessage.setChatId(update.getMessage().getChatId().toString());
            sorrySendMessage.setText(getSorryMessage());
            return sorrySendMessage;
        }
        
        return getCheckInputMessageAndGetAnswer(message);
```

src/main/java/by/custom/utilcalculator/controller/MessageRouter.java
```diff
@@ -0,0 +1,72 @@
...
+            case Command.LESS_3_YEARS_AGE, Command.BETWEEN_3_AND_7_YEARS_AGE, Command.MORE_7_YEARS_AGE ->
+                    sendMessage.setText(botFieldsManager.processCarAge(usersMessage));
+            case Command.GASOLINE_TYPE_ENGINE, Command.ELECTRIC_TYPE_ENGINE ->
+                    sendMessage.setText(botFieldsManager.processEngineType(usersMessage));
+            case Command.VOLUME_LESS_1000_CM, Command.VOLUME_BETWEEN_1000_2000_CM, Command.VOLUME_BETWEEN_2000_3000_CM,
+                    Command.VOLUME_BETWEEN_3000_3500_CM, Command.VOLUME_MORE_3500_CM ->
+                    sendMessage.setText(botFieldsManager.processEngineVolume(usersMessage));
+            default -> sendMessage.setText(getSorryMessage());
+        }
+        return sendMessage;
```
MaximPilipeiko, line 58: I prefer do not split creating object on several lines.
```suggestion
        return new SendMessage(userId, answer);
```

src/main/java/by/custom/utilcalculator/service/BotFieldsManager.java
```diff
@@ -0,0 +1,93 @@
+package by.custom.utilcalculator.service;
+
+import by.custom.utilcalculator.directory.UserProgress;
+import by.custom.utilcalculator.directory.resources.Command;
+import by.custom.utilcalculator.directory.steps.*;
+
+public class BotFieldsManager {
+    private final UserProgress userProgress;
+    private final MessagesCreator messagesManager;
```
MaximPilipeiko, line 9:
```suggestion
    private final MessagesCreator messagesCreator;
```

src/main/java/by/custom/utilcalculator/service/BotFieldsManager.java
```diff
@@ -0,0 +1,93 @@
+package by.custom.utilcalculator.service;
+
+import by.custom.utilcalculator.directory.UserProgress;
+import by.custom.utilcalculator.directory.resources.Command;
+import by.custom.utilcalculator.directory.steps.*;
+
+public class BotFieldsManager {
```
MaximPilipeiko, line 7: It's not call to action
```suggestion
public class UserAnswerManager {
```

src/main/java/by/custom/utilcalculator/controller/MessageRouter.java
```diff
@@ -0,0 +1,72 @@
...
+    private MessageRouter() {
+        messagesCreator = MessagesCreator.getInstance();
+        botFieldsManager = BotFieldsManager.getInstance();
+    }
+
+    public static MessageRouter getInstance() {
+        return MessagesCheckerHolder.MESSAGES_CHECKER;
+    }
+
+    public SendMessage checkMessageBeforeExecutionAndGetResult(Update update) {
```
MaximPilipeiko, line 23: It's not call to action
```suggestion
    public SendMessage route(Update update) {
```

src/main/java/by/custom/utilcalculator/service/BotFieldsManager.java
MaximPilipeiko: You are absolutely right. But finally we need user info to save user progress. But it's another story.

MaximPilipeiko: It's ok for service. It's a right place to aggregate these things, store progress, and prepare answer.

MaximPilipeiko (approved):

src/main/java/by/custom/utilcalculator/controller/MessageRouter.java
Y-Pazniak: Renamed.

src/main/java/by/custom/utilcalculator/service/BotFieldsManager.java
Y-Pazniak: Renamed to UserProgressManager

src/main/java/by/custom/utilcalculator/service/BotFieldsManager.java
Y-Pazniak: Fixed.

src/main/java/by/custom/utilcalculator/controller/MessageRouter.java
Y-Pazniak: Sorry, it is impossible to add userId and answer into SendMessage object - it's constructor is empty. 
We can write a wrapper, but in my opinion it is reduntant. 
I fixed it in other way: 
```
        SendMessage sendMessage = new SendMessage();
        sendMessage.setChatId(userID);
        sendMessage.setText(answer);
        return sendMessage;
```

src/main/java/by/custom/utilcalculator/controller/MessageRouter.java
Y-Pazniak: Yes, it is really shorter than my variant. Fixed, thanks.

PR #16, opened 2023-10-28
It is bad solution, and program doesn't work. I will try to fix it ne…

Y-Pazniak:
…xt week


MaximPilipeiko (changes requested):
Please, ping me, when it will be ready to review

pom.xml
```diff
@@ -38,8 +38,8 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
                 <configuration>
-                    <source>21</source>
-                    <target>21</target>
+                    <source>19</source>
```
MaximPilipeiko: Why downgrade?

src/main/java/by/custom/utilcalculator/controller/MessageRouter.java
```diff
@@ -36,22 +36,24 @@ public SendMessage route(Update update) {
...
 
         switch (usersMessage) {
-            case Command.START -> answer = getGreetingMessage();
+            case Command.START -> {
+                userProgressManager.createNewUserProgress(message);
+                answer = getGreetingMessage();
+            }
             case Command.EAES, Command.OTHER_COUNTRIES ->
-                    answer = botFieldsManager.processCarOrigin(usersMessage, userProgress);
+                    answer = userProgressManager.processCarOrigin(usersMessage, message);
```
MaximPilipeiko: just userId should be enough.
If it's possible it's better that `userProgressManager` don't know about telegram api.
```suggestion
                    answer = userProgressManager.processCarOrigin(userId, usersMessage);
```

src/main/java/by/custom/utilcalculator/domain/UserProgress.java
```diff
@@ -2,19 +2,18 @@
...
-public class UserProgress {
+import java.io.Serializable;
+
+public class UserProgress implements Serializable {
     private CountryOrigin countryOrigin = null;
     private OwnersType ownersType = null;
     private CarAge carAge = null;
     private TypeOfEngine typeOfEngine = null;
     private VolumeOfEngine volumeOfEngine = null;
+    private final String chatID;
```
MaximPilipeiko, line 13: Please, for consistency use something one `chatId` or `userId`

src/main/java/by/custom/utilcalculator/domain/UserProgressStorage.java
```diff
@@ -1,27 +1,52 @@
...
 import java.util.HashMap;
 
 public class UserProgressStorage {
-    private final HashMap<String, UserProgress> users;
+
+    private static class UserProgressStorageHolder {
+        private static final UserProgressStorage USER_PROGRESS_STORAGE = new UserProgressStorage();
+    }
+
+    private HashMap<String, UserProgress> users;
```
MaximPilipeiko: Please, use abstraction instead implementation
```suggestion
    private Map<String, UserProgress> users;
```

src/main/java/by/custom/utilcalculator/domain/UserProgressStorage.java
```diff
@@ -1,27 +1,52 @@
 package by.custom.utilcalculator.domain;
 
+import org.telegram.telegrambots.meta.api.objects.Message;
+
+import java.io.*;
 import java.util.HashMap;
 
 public class UserProgressStorage {
```
MaximPilipeiko: Please, create interface for UserProgressStorage. Could be done in future. Lets discuss it during call.
```suggestion
public class UserProgressStorage {
```

src/main/java/by/custom/utilcalculator/domain/UserProgressStorage.java
MaximPilipeiko: I recommend removing `Map` and storing data only in file(s). I remember that you liked idea of storing every user in separate file. Maybe you could implement it?

src/main/java/by/custom/utilcalculator/service/UserProgressManager.java
```diff
@@ -61,12 +76,15 @@ public String processEngineVolume(String command, UserProgress userProgress) {
                     userProgress.setVolumeOfEngine(VolumeOfEngine.BETWEEN_3000_AND_3500);
             case Command.VOLUME_MORE_3500_CM -> userProgress.setVolumeOfEngine(VolumeOfEngine.MORE_3500);
         }
+        userProgressStorage.save();
         return messagesCreator.getCountryOrigin(userProgress);
     }
 
-
+    private UserProgress getUserProgress(Message message) {
```
MaximPilipeiko: I think this metod is redundant

src/main/java/by/custom/utilcalculator/domain/UserProgressStorage.java
```diff
@@ -1,27 +1,52 @@
...
             users.put(chatId, userProgress);
-            return userProgress;
-        } else return users.get(chatId);
+        }
+        save();
     }
 
-    private static class UserProgressStorageHolder {
-        private static final UserProgressStorage USER_PROGRESS_STORAGE = new UserProgressStorage();
+    public void save() {
```
MaximPilipeiko: It's strange that `save` do not accept any params. What does it save? It's not clear.
```suggestion
    public void save(UserProgress userProgress) {
```

MaximPilipeiko (changes requested):

pom.xml
Y-Pazniak: Sorry, my mistake. The case of error is my inattention.
Fixed.

src/main/java/by/custom/utilcalculator/controller/MessageRouter.java
Y-Pazniak: Fixed.

src/main/java/by/custom/utilcalculator/domain/UserProgress.java
Y-Pazniak: chatId  is better. I renamed all the variables into the same way

src/main/java/by/custom/utilcalculator/domain/UserProgressStorage.java
Y-Pazniak: Sorry, bad habit. Fixed.

src/main/java/by/custom/utilcalculator/domain/UserProgressStorage.java
Y-Pazniak: Realized. Looks great

src/main/java/by/custom/utilcalculator/domain/UserProgressStorage.java
Y-Pazniak: Realized. It was easyer than I thought. I am not sure about the code, but at least it seems to work. I am happy. Thanks for yout support!

src/main/java/by/custom/utilcalculator/service/UserProgressManager.java
Y-Pazniak: Deleted.

src/main/java/by/custom/utilcalculator/domain/UserProgressStorage.java
Y-Pazniak: Fixed.

src/main/java/by/custom/utilcalculator/controller/MessageRouter.java
```diff
@@ -35,28 +35,30 @@ public SendMessage route(Update update) {
...
-                    answer = botFieldsManager.processEngineType(usersMessage, userProgress);
+                    answer = userProgressManager.processEngineType(usersMessage, chatID);
             case Command.VOLUME_LESS_1000_CM, Command.VOLUME_BETWEEN_1000_2000_CM, Command.VOLUME_BETWEEN_2000_3000_CM,
                     Command.VOLUME_BETWEEN_3000_3500_CM, Command.VOLUME_MORE_3500_CM ->
-                    answer = botFieldsManager.processEngineVolume(usersMessage, userProgress);
+                    answer = userProgressManager.processEngineVolume(usersMessage, chatID);
             default -> answer = getSorryMessage();
         }
 
         SendMessage sendMessage = new SendMessage();
```
MaximPilipeiko: I recommend to use appropriate constructor and remove default(without params) constructor. In this case, it's impossible to create "invalid" object, an object with `null` fields. 
```suggestion
        SendMessage sendMessage = new SendMessage(chatId, answer);
```

src/main/java/by/custom/utilcalculator/domain/FileUserProgressStorage.java
```diff
@@ -0,0 +1,60 @@
+package by.custom.utilcalculator.domain;
+
+import by.custom.utilcalculator.exception.UserFileNotFoundException;
+
+import java.io.*;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+
+public class FileUserProgressStorage implements iUserProgressStorage {
+    private final String path = "D:/%s.dat";
```
MaximPilipeiko: Please, move `"D:/"` to env variable. In this case you can run your code on machine where there is no disk "D:/" without code changes.

src/main/java/by/custom/utilcalculator/domain/FileUserProgressStorage.java
```diff
@@ -0,0 +1,60 @@
+package by.custom.utilcalculator.domain;
+
+import by.custom.utilcalculator.exception.UserFileNotFoundException;
+
+import java.io.*;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+
+public class FileUserProgressStorage implements iUserProgressStorage {
```
MaximPilipeiko:
```suggestion
public class FileUserProgressStorage implements  IUserProgressStorage {
```

src/main/java/by/custom/utilcalculator/domain/FileUserProgressStorage.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    public static FileUserProgressStorage getInstance() {
+        return FileUserProgressStorageHolder.FILE_USER_PROGRESS_STORAGE;
+    }
+
+    @Override
+    public void save(UserProgress userProgress) {
+        try {
+            ObjectOutputStream oos = new ObjectOutputStream(new FileOutputStream(String.format(path, userProgress.getChatID())));
```
MaximPilipeiko: What do you think to encapsulate building path to file logic in dedicated method?
```suggestion
            ObjectOutputStream oos = new ObjectOutputStream(new FileOutputStream(buildFilePath(path, userProgress.getChatID())));
```

src/main/java/by/custom/utilcalculator/domain/FileUserProgressStorage.java
```diff
@@ -0,0 +1,60 @@
...
+            oos.writeObject(userProgress);
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    @Override
+    public UserProgress get(String chatID) throws UserFileNotFoundException {
+        UserProgress localUserProgress;
+        if (isUserFileExists(chatID)) {
```
MaximPilipeiko: To avoid redundant inner block there is "fail-fast" approach:
```suggestion
        if (!isUserFileExists(chatID)) {
            throw new UserFileNotFoundException();
        }
        try {
                ObjectInputStream ois = new ObjectInputStream(new FileInputStream(String.format(path, chatID)));
               return (UserProgress) ois.readObject();
            } catch (IOException | ClassNotFoundException e) {
                throw new RuntimeException(e);
            }
```

src/main/java/by/custom/utilcalculator/domain/FileUserProgressStorage.java
```diff
@@ -0,0 +1,60 @@
...
+            try {
+                ObjectInputStream ois = new ObjectInputStream(new FileInputStream(String.format(path, chatID)));
+                localUserProgress = (UserProgress) ois.readObject();
+                return localUserProgress;
+            } catch (IOException | ClassNotFoundException e) {
+                throw new RuntimeException(e);
```
MaximPilipeiko: I recommend using a custom exception. `ReadingUserProgressFromFileException` or smth like this.
And let's try to avoid RuntimeException (it's not a strict rule).

src/main/java/by/custom/utilcalculator/domain/FileUserProgressStorage.java
```diff
@@ -0,0 +1,60 @@
...
+        try {
+            ObjectOutputStream oos = new ObjectOutputStream(new FileOutputStream(String.format(path, userProgress.getChatID())));
+            oos.writeObject(userProgress);
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    @Override
+    public UserProgress get(String chatID) throws UserFileNotFoundException {
```
MaximPilipeiko: And for other methods please use final with params.
```suggestion
    public UserProgress get(final String chatID) throws UserFileNotFoundException {
```

src/main/java/by/custom/utilcalculator/domain/MapUserProgressStorage.java
```diff
@@ -0,0 +1,52 @@
...
+        if (!users.containsKey(chatId)) {
+            UserProgress userProgress = new UserProgress(chatId);
+            users.put(chatId, userProgress);
+            save(userProgress);
+        }
+    }
+
+    @Override
+    public void save(UserProgress userProgress) {
+        try (ObjectOutputStream oos = new ObjectOutputStream(new FileOutputStream("D:/users.dat"))) {
```
MaximPilipeiko: I think, store in file is redundant

src/main/java/by/custom/utilcalculator/domain/iUserProgressStorage.java
```diff
@@ -0,0 +1,9 @@
+package by.custom.utilcalculator.domain;
+
+import by.custom.utilcalculator.exception.UserFileNotFoundException;
+
+public interface iUserProgressStorage {
```
MaximPilipeiko:
```suggestion
public interface IUserProgressStorage {
```

src/main/java/by/custom/utilcalculator/exception/UserFileNotFoundException.java
```diff
@@ -0,0 +1,15 @@
...
+
+public class UserFileNotFoundException extends Exception {
+    private final BundleResourcesServant bundle;
+    public UserFileNotFoundException(){
+        bundle = BundleResourcesServant.getInstance();
+    }
+
+    @Override
+    public String getMessage() {
+        return bundle.getString("answers.user_not_found");
```
MaximPilipeiko: In my view Exceptions should not own bundle 
```suggestion
        return BundleResourcesServant.getInstance().getString("answers.user_not_found");
```

src/main/java/by/custom/utilcalculator/service/UserProgressManager.java
```diff
@@ -1,72 +1,112 @@
...
+import by.custom.utilcalculator.domain.FileUserProgressStorage;
 import by.custom.utilcalculator.domain.UserProgress;
 import by.custom.utilcalculator.domain.constants.Command;
 import by.custom.utilcalculator.domain.constants.steps.*;
+import by.custom.utilcalculator.exception.UserFileNotFoundException;
 
 public class UserProgressManager {
-    //private final UserProgress userProgress;
     private final MessagesCreator messagesCreator;
+    private final FileUserProgressStorage fileUserProgressStorage;
```
MaximPilipeiko: Please, before using any class think "Can I use abstraction?"
```suggestion
    private final IUserProgressStorage userProgressStorage;
```

src/main/java/by/custom/utilcalculator/service/UserProgressManager.java
```diff
@@ -1,72 +1,112 @@
...
+        UserProgress userProgress;
+        try {
+            userProgress = fileUserProgressStorage.get(chatID);
+            switch (command) {
+                case Command.JURIDICAL_PERSON -> userProgress.setOwnersType(OwnersType.JURIDICAL);
+                case Command.PHYSICAL_PERSON -> userProgress.setOwnersType(OwnersType.PHYSICAL);
+            }
+            fileUserProgressStorage.save(userProgress);
+            return messagesCreator.getCountryOrigin(userProgress);
+        } catch (UserFileNotFoundException e) {
```
MaximPilipeiko: I think it's better to move catch logic to the router. but this is a controversial decision in any case. Please, do as you feel in this case.

MaximPilipeiko (changes requested):

src/main/java/by/custom/utilcalculator/controller/MessageRouter.java
Y-Pazniak: The brilliant idea, i like it. Done.

src/main/java/by/custom/utilcalculator/domain/FileUserProgressStorage.java
Y-Pazniak: Done.

src/main/java/by/custom/utilcalculator/domain/FileUserProgressStorage.java
Y-Pazniak: Renamed.

src/main/java/by/custom/utilcalculator/domain/FileUserProgressStorage.java
Y-Pazniak: Good point. Done.

src/main/java/by/custom/utilcalculator/domain/FileUserProgressStorage.java
Y-Pazniak: Done.

src/main/java/by/custom/utilcalculator/domain/FileUserProgressStorage.java
Y-Pazniak: New excertion was added.

src/main/java/by/custom/utilcalculator/domain/FileUserProgressStorage.java
Y-Pazniak: OK. Let's discuss, why.

src/main/java/by/custom/utilcalculator/domain/MapUserProgressStorage.java
Y-Pazniak: Haven't done it yet. 
Are you sure we need to fix MapUserProgressStorage? I used this file to train skill of saving into file. We can try to make smth meaningful from this class, but it is like an outdated draft, in my honest opinion.

src/main/java/by/custom/utilcalculator/domain/iUserProgressStorage.java
Y-Pazniak: Fixed

src/main/java/by/custom/utilcalculator/exception/UserFileNotFoundException.java
Y-Pazniak: I couldn't even imagine this construction. Sure, now we can remove bundle and use new construstion

src/main/java/by/custom/utilcalculator/service/UserProgressManager.java
Y-Pazniak: OK, thats a great trick. Before creating an object ask myself "is it possible to use abstr.?"
Thank you

src/main/java/by/custom/utilcalculator/service/UserProgressManager.java
Y-Pazniak: It is as broad as it is long. I don't like both of this ways - they made from our pretty code some kind of spagetti-code.

src/main/java/by/custom/utilcalculator/controller/MessageRouter.java
```diff
@@ -33,32 +35,33 @@ public SendMessage route(Update update) {
...
+                        answer = userProgressManager.processCarAge(usersMessage, chatID);
+                case Command.GASOLINE_TYPE_ENGINE, Command.ELECTRIC_TYPE_ENGINE ->
+                        answer = userProgressManager.processEngineType(usersMessage, chatID);
+                case Command.VOLUME_LESS_1000_CM, Command.VOLUME_BETWEEN_1000_2000_CM, Command.VOLUME_BETWEEN_2000_3000_CM,
+                        Command.VOLUME_BETWEEN_3000_3500_CM, Command.VOLUME_MORE_3500_CM ->
+                        answer = userProgressManager.processEngineVolume(usersMessage, chatID);
+                default -> answer = getSorryMessage();
+            }
+        } catch (UserFileNotFoundException e) {
+            return new CustomSendMessage(chatID, e.getMessage());
```
MaximPilipeiko: Thibk about to translate it here. Exception contains only error code.

src/main/java/by/custom/utilcalculator/domain/FileUserProgressStorage.java
```diff
@@ -0,0 +1,68 @@
...
+
+import by.custom.utilcalculator.exception.ReadingUserProgressFromFileException;
+import by.custom.utilcalculator.exception.UserFileNotFoundException;
+
+import java.io.*;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+
+public class FileUserProgressStorage implements IUserProgressStorage {
+    private final String databaseEnvVariable = System.getenv("CHATBOT_DATABASE_STORAGE");
```
MaximPilipeiko: I think it corresponds to trues more.
```suggestion
    private final String databaseEnvVariable = System.getenv("CHATBOT_FILE_STORAGE");
```

src/main/java/by/custom/utilcalculator/domain/FileUserProgressStorage.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    public static FileUserProgressStorage getInstance() {
+        return FileUserProgressStorageHolder.FILE_USER_PROGRESS_STORAGE;
+    }
+
+    @Override
+    public void save(final UserProgress userProgress) {
+        try {
+            ObjectOutputStream oos = new ObjectOutputStream(new FileOutputStream(buildFilePath(userProgress.getChatID())));
```
MaximPilipeiko: Please, use [try-with-resources](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html)

src/main/java/by/custom/utilcalculator/domain/FileUserProgressStorage.java
```diff
@@ -0,0 +1,68 @@
...
+            oos.writeObject(userProgress);
+        } catch (IOException e) {
+            throw new RuntimeException(e);
```
MaximPilipeiko: It's better to use your custom exception. For example `WritingUserProgressToFileException`

src/main/java/by/custom/utilcalculator/domain/FileUserProgressStorage.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    @Override
+    public UserProgress get(final String chatID) throws UserFileNotFoundException, ReadingUserProgressFromFileException {
+        if (!isUserFileExists(chatID)) {
+            throw new UserFileNotFoundException();
+        }
+        ObjectInputStream ois;
+        try {
+            ois = new ObjectInputStream(new FileInputStream(buildFilePath(chatID)));
```
MaximPilipeiko: Please, use try-with-resources.

src/main/java/by/custom/utilcalculator/domain/FileUserProgressStorage.java
```diff
@@ -0,0 +1,68 @@
...
+            return (UserProgress) ois.readObject();
+        } catch (IOException | ClassNotFoundException e) {
+            e.printStackTrace();
+            throw new ReadingUserProgressFromFileException();
+        }
+    }
+
+    public void createNewUser (final String chatID) {
```
MaximPilipeiko: In mu view, just `create` will more correlate with other method `save`, `get`.
```suggestion
    public void create(final String chatID) {
```

src/main/java/by/custom/utilcalculator/domain/MapUserProgressStorage.java
```diff
@@ -0,0 +1,52 @@
...
+public class MapUserProgressStorage implements IUserProgressStorage {
+    public static MapUserProgressStorage getInstance() {
+        return UserProgressStorageHolder.USER_PROGRESS_STORAGE;
+    }
+
+    private static class UserProgressStorageHolder {
+        private static final MapUserProgressStorage USER_PROGRESS_STORAGE = new MapUserProgressStorage();
+    }
+
+    private Map<String, UserProgress> users;
```
MaximPilipeiko:
```suggestion
    private final Map<String, UserProgress> users;
```

src/main/java/by/custom/utilcalculator/domain/MapUserProgressStorage.java
MaximPilipeiko: Let's discuss it on call. If it's just for the train, it's ok. But I prefer to keep the codebase as pure as possible.

src/main/java/by/custom/utilcalculator/exception/UserFileNotFoundException.java
```diff
@@ -0,0 +1,10 @@
+package by.custom.utilcalculator.exception;
+
+import by.custom.utilcalculator.service.BundleResourcesServant;
+
+public class UserFileNotFoundException extends Exception {
```
MaximPilipeiko: I think, it might be helpful if with and other exception contains more info. Like `file path`.

src/main/java/by/custom/utilcalculator/service/CustomSendMessage.java
```diff
@@ -0,0 +1,11 @@
+package by.custom.utilcalculator.service;
+
+import org.telegram.telegrambots.meta.api.methods.send.SendMessage;
+
+public class CustomSendMessage extends SendMessage {
+
+    public CustomSendMessage(final String chatID, final String answer) {
+        super.setChatId(chatID);
```
MaximPilipeiko: maybe there is constructor `super(chatID, answer)`?

MaximPilipeiko (changes requested):

src/main/java/by/custom/utilcalculator/domain/MapUserProgressStorage.java
Y-Pazniak: Fixed.

src/main/java/by/custom/utilcalculator/controller/MessageRouter.java
Y-Pazniak: Error code was added

src/main/java/by/custom/utilcalculator/domain/FileUserProgressStorage.java
Y-Pazniak: Yes, fixed

src/main/java/by/custom/utilcalculator/domain/FileUserProgressStorage.java
Y-Pazniak: Done, thanks for reminding about this construction

src/main/java/by/custom/utilcalculator/domain/FileUserProgressStorage.java
Y-Pazniak: Done

src/main/java/by/custom/utilcalculator/domain/FileUserProgressStorage.java
Y-Pazniak: Done

src/main/java/by/custom/utilcalculator/domain/FileUserProgressStorage.java
Y-Pazniak: Renamed

src/main/java/by/custom/utilcalculator/domain/MapUserProgressStorage.java
Y-Pazniak: Fixed

src/main/java/by/custom/utilcalculator/exception/UserFileNotFoundException.java
Y-Pazniak: I tried to add more info into exceptions, but I am not sure about success.

src/main/java/by/custom/utilcalculator/service/CustomSendMessage.java
Y-Pazniak: Yes, this construction really was there and I could use it without a custom wrapper. Thank you for advice, fixed

src/main/java/by/custom/utilcalculator/exception/UtilsborException.java
```diff
@@ -0,0 +1,20 @@
+package by.custom.utilcalculator.exception;
+
+import org.telegram.telegrambots.meta.api.objects.Message;
+
+public abstract class UtilsborException extends Exception {
+    public static final String USER_FILE_NOT_FOUND = "user_file_not_found";
```
MaximPilipeiko: You could move it to enum `UtilsborErrorCode`

src/main/java/by/custom/utilcalculator/exception/UtilsborException.java
```diff
@@ -0,0 +1,20 @@
...
+    public static final String READING_FROM_FILE_EXCEPTION = "file_reading_error";
+    public static final String SAVING_INTO_FILE_EXCEPTION = "file_saving_error";
+    public abstract String getExceptionCode();
+
+    public abstract String getFileNameExceptionDetails(final String filename);
```
MaximPilipeiko: What if exception is not connected with file. Database exception for example. I think this method is not needed

src/main/java/by/custom/utilcalculator/exception/UtilsborException.java
```diff
@@ -0,0 +1,20 @@
...
+
+    public abstract String getUserChatIDExceptionDetails(final String chatID);
+
+    public abstract String getUserCommandExceptionDetails(final String userCommand);
+
+    public String[] getMessageInfo(final Message message) {
```
MaximPilipeiko: Message info is redundant too

src/main/java/by/custom/utilcalculator/exception/UtilsborException.java
```diff
@@ -0,0 +1,20 @@
...
+
+public abstract class UtilsborException extends Exception {
+    public static final String USER_FILE_NOT_FOUND = "user_file_not_found";
+    public static final String READING_FROM_FILE_EXCEPTION = "file_reading_error";
+    public static final String SAVING_INTO_FILE_EXCEPTION = "file_saving_error";
+    public abstract String getExceptionCode();
+
+    public abstract String getFileNameExceptionDetails(final String filename);
+
+    public abstract String getUserChatIDExceptionDetails(final String chatID);
```
MaximPilipeiko: Even this method is not required.

src/main/java/by/custom/utilcalculator/exception/UtilsborException.java
```diff
@@ -0,0 +1,20 @@
...
+
+    public abstract String getUserCommandExceptionDetails(final String userCommand);
```
MaximPilipeiko: And this. You can move directly to children class

MaximPilipeiko (changes requested):

src/main/java/by/custom/utilcalculator/controller/MessageRouter.java
```diff
@@ -33,32 +37,48 @@ public SendMessage route(Update update) {
...
-    private SendMessage getCheckInputMessageAndGetAnswer(Message message) {
+    private SendMessage getCheckInputMessageAndGetAnswer(final Message message) {
         String usersMessage = message.getText();
-        String userID = message.getChatId().toString();
-        UserProgress userProgress = UserStorageManager.getInstance().getUserProgress(userID);
+        String chatID = message.getChatId().toString();
         String answer;
+        try {
+            answer = route(usersMessage, chatID);
+        } catch (UserFileNotFoundException | ReadingUserProgressFromFileException |
```
MaximPilipeiko:
```suggestion
        } catch (UtilsborException e)
```

src/main/java/by/custom/utilcalculator/exception/UtilsborException.java
Y-Pazniak: Done.

src/main/java/by/custom/utilcalculator/exception/UtilsborException.java
Y-Pazniak: Removed.

src/main/java/by/custom/utilcalculator/exception/UtilsborException.java
Y-Pazniak: Deleted

src/main/java/by/custom/utilcalculator/exception/UtilsborException.java
Y-Pazniak: Deleted

src/main/java/by/custom/utilcalculator/exception/UtilsborException.java
Y-Pazniak: Deleted

src/main/java/by/custom/utilcalculator/controller/Bot.java
```diff
@@ -22,6 +23,7 @@ public static Bot getBot() {
     //private constructor to avoid wrong bot's creation
     private Bot() {
         messageRouter = MessageRouter.getInstance();
+        initDB();
```
MaximPilipeiko, line 25: it must be the first

src/main/java/by/custom/utilcalculator/domain/PostgresUserProgress.java
```diff
@@ -0,0 +1,22 @@
+package by.custom.utilcalculator.domain;
+
+import java.sql.*;
+
+public class PostgresUserProgress {
+
+    public void testDatabase() {
+        String jdbcURL = "jdbc:postgresql://localhost:5432/test";
```
MaximPilipeiko: Please, take this param from env variables

src/main/java/by/custom/utilcalculator/domain/IUserProgressStorage.java
```diff
@@ -0,0 +1,15 @@
...
+import by.custom.utilcalculator.exception.WritingUserProgressIntoFileException;
+
+public interface IUserProgressStorage {
+    void save(final UserProgress userProgress) throws WritingUserProgressIntoFileException;
+
+    UserProgress get(final String chatID) throws UserFileNotFoundException, ReadingUserProgressFromFileException;
+
+    void create(final String chatID) throws WritingUserProgressIntoFileException;
+
+    String getPath(final String chatID);
```
MaximPilipeiko: I suggest this method should not be there, because it's specific for FileStorage

src/main/java/by/custom/utilcalculator/domain/FileUserProgressStorage.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    @Override
+    public void save(final UserProgress userProgress) throws WritingUserProgressIntoFileException {
+        String chatID = userProgress.getChatID();
+        String filePath = buildFilePath(chatID);
+        try (ObjectOutputStream oos = new ObjectOutputStream(new FileOutputStream(filePath))){
+            oos.writeObject(userProgress);
+        } catch (IOException e) {
+            throw new WritingUserProgressIntoFileException(chatID, filePath);
```
MaximPilipeiko: Please, add the cause IOException to the constructor as well

src/main/java/by/custom/utilcalculator/domain/FileUserProgressStorage.java
```diff
@@ -0,0 +1,72 @@
...
+    @Override
+    public UserProgress get(final String chatID) throws UserFileNotFoundException, ReadingUserProgressFromFileException {
+        String filePath = buildFilePath(chatID);
+        if (!isUserFileExists(chatID)) {
+            throw new UserFileNotFoundException(chatID, filePath);
+        }
+        try (ObjectInputStream ois  = new ObjectInputStream(new FileInputStream(filePath))){
+            return (UserProgress) ois.readObject();
+        } catch (IOException | ClassNotFoundException e) {
+            throw new ReadingUserProgressFromFileException(chatID, filePath);
```
MaximPilipeiko: Please, add the cause Exception to the constructor as well

src/main/java/by/custom/utilcalculator/domain/FileUserProgressStorage.java
```diff
@@ -0,0 +1,72 @@
...
+        }
+    }
+
+    public void create(final String chatID) throws WritingUserProgressIntoFileException {
```
MaximPilipeiko: I think this method is redundant. `save` is enough

src/main/java/by/custom/utilcalculator/domain/MapUserProgressStorage.java
```diff
@@ -0,0 +1,42 @@
+package by.custom.utilcalculator.domain;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class MapUserProgressStorage implements IUserProgressStorage {
```
MaximPilipeiko:
```suggestion
public class MemoryUserProgressStorage implements IUserProgressStorage {
```

src/main/java/by/custom/utilcalculator/exception/UtilsborException.java
```diff
@@ -0,0 +1,22 @@
+package by.custom.utilcalculator.exception;
+
+public abstract class UtilsborException extends Exception {
+    private final String errorCode;
+
+    public UtilsborException(final String errorCode) {
+        this.errorCode = errorCode;
+    }
+
+    public UtilsborException(final String errorCode, final String detailsAboutError) {
```
MaximPilipeiko:
```suggestion
    public UtilsborException(final String errorCode, final String errorDetails) {
```

src/main/java/by/custom/utilcalculator/service/UserProgressManager.java
```diff
@@ -1,72 +1,98 @@
...
 import by.custom.utilcalculator.domain.constants.Command;
 import by.custom.utilcalculator.domain.constants.steps.*;
+import by.custom.utilcalculator.exception.ReadingUserProgressFromFileException;
+import by.custom.utilcalculator.exception.UserFileNotFoundException;
+import by.custom.utilcalculator.exception.WritingUserProgressIntoFileException;
 
 public class UserProgressManager {
-    //private final UserProgress userProgress;
     private final MessagesCreator messagesCreator;
+    private final IUserProgressStorage fileUserProgressStorage;
```
MaximPilipeiko:
```suggestion
    private final IUserProgressStorage userProgressStorage;
```

src/main/java/by/custom/utilcalculator/service/UserProgressManager.java
```diff
@@ -1,72 +1,98 @@
...
     }
 
-    public String processCarAge(String command, UserProgress userProgress) {
-        switch (command) {
-            case Command.LESS_3_YEARS_AGE -> userProgress.setCarAge(CarAge.LESS_3_YEARS);
-            case Command.BETWEEN_3_AND_7_YEARS_AGE -> userProgress.setCarAge(CarAge.BETWEEN_3_AND_7_YEARS);
-            case Command.MORE_7_YEARS_AGE -> userProgress.setCarAge(CarAge.MORE_7_YEARS);
-        }
-        return messagesCreator.getCountryOrigin(userProgress);
+    public String processOwnerType(final String command, final String chatID) throws UserFileNotFoundException, ReadingUserProgressFromFileException, WritingUserProgressIntoFileException {
```
MaximPilipeiko:
```suggestion
    public String processOwnerType(final String command, final String chatID) throws UtilsborException {
```

src/main/java/by/custom/utilcalculator/service/UserProgressManager.java
```diff
@@ -1,72 +1,98 @@
...
+        UserProgress userProgress;
+            userProgress = fileUserProgressStorage.get(chatID);
+            switch (command) {
+                case Command.JURIDICAL_PERSON -> userProgress.setOwnersType(OwnersType.JURIDICAL);
+                case Command.PHYSICAL_PERSON -> userProgress.setOwnersType(OwnersType.PHYSICAL);
+            }
+            fileUserProgressStorage.save(userProgress);
+            return messagesCreator.getCountryOrigin(userProgress);
```
MaximPilipeiko: Are you sure that `getCountryOrigin` should be there?

MaximPilipeiko (changes requested):
Good. Just few questions

src/main/java/by/custom/utilcalculator/controller/MessageRouter.java
Y-Pazniak: Fixed

src/main/java/by/custom/utilcalculator/controller/Bot.java
Y-Pazniak: removed

src/main/java/by/custom/utilcalculator/domain/PostgresUserProgress.java
Y-Pazniak: Done

src/main/java/by/custom/utilcalculator/domain/IUserProgressStorage.java
Y-Pazniak: Deleted

src/main/java/by/custom/utilcalculator/domain/FileUserProgressStorage.java
Y-Pazniak: Great advice, done

src/main/java/by/custom/utilcalculator/domain/FileUserProgressStorage.java
Y-Pazniak: Done, thanks

src/main/java/by/custom/utilcalculator/domain/FileUserProgressStorage.java
Y-Pazniak: Fixed

src/main/java/by/custom/utilcalculator/domain/MapUserProgressStorage.java
Y-Pazniak: Renamed

src/main/java/by/custom/utilcalculator/exception/UtilsborException.java
Y-Pazniak: Renamed

src/main/java/by/custom/utilcalculator/service/UserProgressManager.java
Y-Pazniak: Renamed

src/main/java/by/custom/utilcalculator/service/UserProgressManager.java
Y-Pazniak: Done, sorry for this mistake

src/main/java/by/custom/utilcalculator/service/UserProgressManager.java
Y-Pazniak: From this method starts to build string with a user choise. I have added a wrapper-method to imrove this code. It works fine, but naming was really confusing

src/main/java/by/custom/utilcalculator/domain/FileUserProgressStorage.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    @Override
+    public void save(final UserProgress userProgress) throws WritingUserProgressIntoFileException {
+        String chatID = userProgress.getChatID();
+        String filePath = buildFilePath(chatID);
+        try (ObjectOutputStream oos = new ObjectOutputStream(new FileOutputStream(filePath))){
+            oos.writeObject(userProgress);
+        } catch (IOException e) {
+            throw new WritingUserProgressIntoFileException(chatID, filePath, e.getCause());
```
MaximPilipeiko: The cause of `WritingUserProgressIntoFileException` is `IOException` and the cause of IOException is e.getCause(). We must not lose info of IOException.
```suggestion
            throw new WritingUserProgressIntoFileException(chatID, filePath, e);
```

src/main/java/by/custom/utilcalculator/domain/MemoryUserProgressStorage.java
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    private static class UserProgressStorageHolder {
+        private static final MemoryUserProgressStorage USER_PROGRESS_STORAGE = new MemoryUserProgressStorage();
+    }
+
+    private final Map<String, UserProgress> users;
+
+    private MemoryUserProgressStorage() {
+        users = new HashMap<>();
```
MaximPilipeiko: It's better to use cuncurrent collection here. Just to think. It's not call to action.

src/main/java/by/custom/utilcalculator/service/MessagesCreator.java
```diff
@@ -192,19 +191,23 @@ private String checkForGasolineAuoAge(UserProgress userProgress) {
...
         StringBuilder stringBuilder = new StringBuilder();
         for (String string : strings) {
             stringBuilder.append(string);
         }
         return stringBuilder.toString();
     }
 
-    private String trimFirstAndLastLetters(String toTrim) {
+    public String getSummaryAnswer(final UserProgress userProgress) {
+        return getCountryOrigin(userProgress);
```
MaximPilipeiko, line 207: We have to discuss it. For me looks strange that `getSummaryAnswer` returns only `getCountryOrigin`.

MaximPilipeiko (changes requested):

src/main/java/by/custom/utilcalculator/domain/FileUserProgressStorage.java
Y-Pazniak: It confuses me, but I think, I got it. 
It helps us to get this kind of message: https://c2n.me/4jXbK1z 
If we send just e.getCause() - we have message in stacktrace without Caused by:

src/main/java/by/custom/utilcalculator/domain/MemoryUserProgressStorage.java
Y-Pazniak: What to think about here? We suppose thousands of users - we have to do map cuncurrent.

src/main/java/by/custom/utilcalculator/service/MessagesCreator.java
Y-Pazniak: Yes, it really looks wrong. But I don't know ho to do it better. 
The logic of building message for user works from getCountryOrigin method. It starts to work here: https://c2n.me/4jXc5Uk - and runs until the message completed. 
Probably, we should rename getCountryOrigin into "getMessageForUser" or something like that? 
I guess, it is more naming problem.

PR #18, opened 2024-01-10
Logic of answers creation has been rewritten. 

Y-Pazniak:
There are less confusing methds now.
buildNextStepQuestion() method has been added to the MessagesCreator class - it creates question for user according to the user's progress.
 getNextStep() method has been added to the UserProgress class - it returns NEXT type of question for user.

### DIFF
--- a/src/main/java/by/custom/utilcalculator/domain/constants/steps/CurrentQuestion.java
+++ b/src/main/java/by/custom/utilcalculator/domain/constants/steps/CurrentQuestion.java
@@ -1,0 +1,5 @@
+package by.custom.utilcalculator.domain.constants.steps;
+
+public enum CurrentQuestion {
+    COUNTRY_ORIGIN, OWNERS_TYPE, TYPE_OF_ENGINE, VOLUME_OF_ENGINE, CAR_AGE, FAREWELL
+}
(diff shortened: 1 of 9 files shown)


src/main/java/by/custom/utilcalculator/domain/constants/steps/CurrentQuestion.java
```diff
@@ -0,0 +1,5 @@
+package by.custom.utilcalculator.domain.constants.steps;
+
+public enum CurrentQuestion {
```
MaximPilipeiko: What if we need to store a previous question (step)? We have to create one more enum with name   `PreviousQuestion`? I recommend using more abstract name, just `Question` or `Step`.

src/main/java/by/custom/utilcalculator/service/MessagesCreator.java
```diff
@@ -18,6 +18,32 @@ public static MessagesCreator getInstance() {
         return MessagesCreatorHolder.MESSAGES_CREATOR;
     }
 
+    //this method builds next questions for user to interact with chatbot
+    public String buildNextStepQuestion(final UserProgress userProgress) {
+        CurrentQuestion currentQuestion = userProgress.getNextStep();
```
MaximPilipeiko:
```suggestion
        CurrentQuestion question = userProgress.getNextStep();
```

MaximPilipeiko (commented):
For me `Step` is more suitable name for `CurrentQuestion`.

MaximPilipeiko (approved):

PR #20, opened 2024-04-23
JSON loading had been added

Y-Pazniak:
Also i've added parents filling from json
I've added json loading.


src/main/java/by/custom/utilcalculator/domain/tree/HelperTree.java
```diff
@@ -31,39 +36,58 @@ public static Map<StepsIndicator, String> fillFieldsToCommandsMap() {
...
 
-        //other physical nodes queue
-        Node other = new Node(treeRoot, Command.OTHER_COUNTRIES);
-        treeRoot.addChildren(Arrays.asList(eaes, other));
-        Node otherPhysical = new Node(other, Command.PHYSICAL_PERSON);
-        Node otherPhysicalAge = new Node(otherPhysical, Command.AGE);
-        otherPhysical.addChildren(List.of(otherPhysicalAge));
+        BufferedReader bufferedReader = null;
+        try {
+            bufferedReader = new BufferedReader(new FileReader("/home/eugene/IdeaProjects/custom_calculator_chatbot/src/main/resources/tree.json"));
```
MaximPilipeiko, line 74: Please, use try-with-resources

pom.xml
```diff
@@ -63,6 +63,12 @@
             <artifactId>postgresql</artifactId>
             <version>42.7.0</version>
         </dependency>
+
+        <dependency>
+            <groupId>com.google.code.gson</groupId>
+            <artifactId>gson</artifactId>
```
MaximPilipeiko: Not a call to action, just a notion. Jackson is a little bit more popular than gson and is used in popular frameworks like Spring

src/main/java/by/custom/utilcalculator/service/BundleResourcesServant.java
```diff
@@ -28,11 +28,11 @@ public String getString(final String stringToGetFromBundle) {
 
     private void createBundleResources() {
         try {
-            File fileResources = new File("D:\\IdeaProjects\\custom_calculator_bot\\src\\main\\resources");
-            URL[] resourcesFileUrls = {fileResources.toURI().toURL()};
-            ClassLoader loader = new URLClassLoader(resourcesFileUrls);
+            final File fileResources = new File("D:\\IdeaProjects\\custom_calculator_bot\\src\\main\\resources");
```
MaximPilipeiko, line 31: Using an absolute path is a bad practice. Please, use relevant path like.
this topic could help you https://stackoverflow.com/questions/17351043/how-to-get-absolute-path-to-file-in-resources-folder-of-your-project

src/main/resources/tree.json
```diff
@@ -0,0 +1,85 @@
+{
+  "parent": null,
```
MaximPilipeiko, line 2: you can rid "parent": null. If field doesn't exist, it's null.

MaximPilipeiko (changes requested):

src/main/java/by/custom/utilcalculator/domain/tree/HelperTree.java
```diff
@@ -31,39 +36,35 @@ public static Map<StepsIndicator, String> fillFieldsToCommandsMap() {
...
+            e.printStackTrace();
+        }
 
-        //other physical nodes queue
-        Node other = new Node(treeRoot, Command.OTHER_COUNTRIES);
-        treeRoot.addChildren(Arrays.asList(eaes, other));
-        Node otherPhysical = new Node(other, Command.PHYSICAL_PERSON);
-        Node otherPhysicalAge = new Node(otherPhysical, Command.AGE);
-        otherPhysical.addChildren(List.of(otherPhysicalAge));
+        final ObjectMapper objectMapper = new ObjectMapper();
```
MaximPilipeiko: It could be a static class field. You haven't to create it on every method call.

src/main/java/by/custom/utilcalculator/domain/tree/HelperTree.java
```diff
@@ -31,39 +36,35 @@ public static Map<StepsIndicator, String> fillFieldsToCommandsMap() {
...
-        Node otherJuridical = new Node(other, Command.JURIDICAL_PERSON);
-        other.addChildren(Arrays.asList(otherPhysical, otherJuridical));
-        Node otherElectricEngine = new Node(otherJuridical, Command.ELECTRIC_TYPE_ENGINE);
-        Node otherElectricAge = new Node(otherElectricEngine, Command.AGE);
-        otherElectricEngine.addChildren(List.of(otherElectricAge));
+        Node treeRootJson = null;
+        try {
+            treeRootJson = objectMapper.readValue(file, Node.class);
+        } catch (final IOException e) {
+            e.printStackTrace();
```
MaximPilipeiko: I think we have to stop the application if cannot read the tree. Please, throw exception on higher level.

src/main/resources/tree.json
```diff
@@ -0,0 +1,70 @@
...
+  "key": "/start",
+  "children": [
+    {
+      "key": "/eaes",
+      "children": [
+        {
+          "key": "/physical",
+          "children": [
+            {
+              "key": "age",
```
MaximPilipeiko: Why this key value without `/`

MaximPilipeiko (changes requested):
Lets disscuss deserialisation on a call

src/main/java/by/custom/utilcalculator/controller/Bot.java
```diff
@@ -35,7 +37,8 @@ public String getBotUsername() {
     @Override
     public String getBotToken() {
         return
-                System.getenv("custom_by_utilsbor_bot");
+                "6490861114:AAHVFOpQANIn_EQr1A1PcLrCMNzVlGkJIig";
```
MaximPilipeiko: Why do you hardcode value?

src/main/java/by/custom/utilcalculator/controller/MessageRouter.java
```diff
@@ -35,38 +37,60 @@ public SendMessage route(final Update update) {
...
         return new SendMessage(chatID, answer);
     }
 
     private String route(final String usersMessage, final String chatID) throws UtilsborException {
-        String answer;
-        switch (usersMessage) {
-            case Command.START -> {
+        final String answer;
+
+        while (true) {
```
MaximPilipeiko: Why infinite loop is needed?

src/main/java/by/custom/utilcalculator/controller/MessageRouter.java
```diff
@@ -35,38 +37,60 @@ public SendMessage route(final Update update) {
...
+            if (Objects.equals(usersMessage, Command.START.getCommand())) {
```
MaximPilipeiko: From my view, an infinite loop looks more stylish.

src/main/java/by/custom/utilcalculator/domain/tree/HelperTree.java
```diff
@@ -2,68 +2,73 @@
...
+import java.net.URL;
+import java.nio.file.Paths;
 import java.util.*;
 
 public class HelperTree {
+    private static final ObjectMapper mapper = new ObjectMapper();
+
     public static Map<StepsIndicator, String> fillFieldsToCommandsMap() {
-        Map<StepsIndicator, String> fieldsToCommands = new HashMap<>(14);
+        final Map<StepsIndicator, String> fieldsToCommands = new HashMap<>(14);
```
MaximPilipeiko: Is it possible to use Command as a value instead of String?
```suggestion
        final Map<StepsIndicator, Command> fieldsToCommands = new HashMap<>(14);
```

src/main/java/by/custom/utilcalculator/domain/tree/HelperTree.java
```diff
@@ -2,68 +2,73 @@
...
-        eaesPhysical.addChildren(List.of(eaesPhysicalAge));
-        Node eaesJuridicalAge = new Node(eaesJuridical, Command.AGE);
-        eaesJuridical.addChildren(List.of(eaesJuridicalAge));
+    public static Node buildTree() throws TreeReadingException {
+        URL res = HelperTree.class.getClassLoader().getResource("tree.json");
+        File file = null;
+        try {
+            file = Paths.get(res.toURI()).toFile();
+        } catch (URISyntaxException e) {
+            e.printStackTrace();
```
MaximPilipeiko, line 52: Please throw `TreeReadingException`. It's impossible to handle this exception gracefully.

src/main/java/by/custom/utilcalculator/domain/tree/Node.java
```diff
@@ -1,14 +1,19 @@
...
+import com.fasterxml.jackson.databind.JsonNode;
+
 import java.util.ArrayList;
 import java.util.List;
 
 public class Node {
-    private final Node parent;
+    private Node parent;
     private final List<Node> children;
     private final String key;
```
MaximPilipeiko: It's better to use enum instead of String.
You can look [deserialization node example](https://github.com/MaximPilipeiko/jackson/blob/master/src/main/java/org/example/Main.java)

src/main/java/by/custom/utilcalculator/exception/TreeReadingException.java
```diff
@@ -0,0 +1,12 @@
+package by.custom.utilcalculator.exception;
+
+import by.custom.utilcalculator.exception.constants.UtilsborErrorCode;
+import by.custom.utilcalculator.exception.constants.UtilsborErrorDescription;
+
+import java.io.IOException;
+
+public class TreeReadingException extends IOException {
```
MaximPilipeiko: This name gives more information: what tree, is it our service exception or pure java exception?
```suggestion
public class UrilsborCommandTreeReadingException extends IOException {
```

src/main/java/by/custom/utilcalculator/service/UserProgressManager.java
```diff
@@ -33,93 +35,123 @@ public void createNewUserProgress(final String chatID) throws UtilsborException
...
+        final UserProgress userProgress;
         userProgress = userProgressStorage.get(chatID);
-        String message;
-        switch (requestingCommand) {
-            case Command.EAES -> userProgress.setCountryOrigin(CountryOrigin.EAES);
-            case Command.OTHER_COUNTRIES -> userProgress.setCountryOrigin(CountryOrigin.OTHER);
-
+        final String message;
+
+        if (Objects.equals(requestingCommand, Command.EAES.getCommand())) {
```
MaximPilipeiko: It's betteruse type `Command` for the argument `requestingCommand`  instead of casting `Command` to `String`.

src/main/java/by/custom/utilcalculator/service/UserProgressManager.java
```diff
@@ -33,93 +35,123 @@ public void createNewUserProgress(final String chatID) throws UtilsborException
...
+            userProgress.setCountryOrigin(CountryOrigin.EAES);
+        } else {
```
MaximPilipeiko:
```suggestion
        } else if (Objects.equals(requestingCommand, Command.OTHER_COUNTRIES.getCommand())) {
            userProgress.setCountryOrigin(CountryOrigin.OTHER);
        }
```

src/main/java/by/custom/utilcalculator/service/UserProgressManager.java
```diff
@@ -33,93 +35,123 @@ public void createNewUserProgress(final String chatID) throws UtilsborException
...
-            case Command.VOLUME_BETWEEN_1000_2000_CM ->
-                    userProgress.setVolumeOfEngine(VolumeOfEngine.BETWEEN_1000_AND_2000);
-            case Command.VOLUME_BETWEEN_2000_3000_CM ->
-                    userProgress.setVolumeOfEngine(VolumeOfEngine.BETWEEN_2000_AND_3000);
-            case Command.VOLUME_BETWEEN_3000_3500_CM ->
-                    userProgress.setVolumeOfEngine(VolumeOfEngine.BETWEEN_3000_AND_3500);
-            case Command.VOLUME_MORE_3500_CM -> userProgress.setVolumeOfEngine(VolumeOfEngine.MORE_3500);
+        final String message;
+
+        while (true) {
```
MaximPilipeiko: switch case is better. And why loop is needed is not clear.

src/main/resources/words_ru.properties
```diff
@@ -37,3 +37,5 @@ answers.file_reading_error = Извините, произошла ошибка 
 answers.file_saving_error = Извините, произошла ошибка сохранения данных. Пожалуйста, попробуйте начать с команды /start или свяжитесь с разработчиками
 
 answers.invalid_order_exception = Извините, недостаточно данных для расчета или нарушена очередность. Попробуйте начать с команды /start или свяжитесь с разработчиками
+
+answers.tree_reading_error = impossible to read the tree
```
MaximPilipeiko: just 'tree' is too abstract
```suggestion
answers.tree_reading_error = impossible to read the command tree 
```

MaximPilipeiko (changes requested):

src/main/java/by/custom/utilcalculator/controller/Bot.java
Y-Pazniak: Another solutions don't work on my Ubuntu system by unknown reason.
I use ubuntu documentation to set EnvVariable:
Terminal: export BOT=TESTBOT
Checking: echo $BOT 
Everything is fine. 

Then I follow Java documentation: use System.getenv("MY_VARIABLE"); to take variable.
And it returns null. 

I don't know why it happens and how to fix it 

image

image

src/main/java/by/custom/utilcalculator/controller/MessageRouter.java
Y-Pazniak: We had switch-case here before. 
Now it becomes impossible to use switch-case, so I replaced it by if.
If i use if-else, it is a terrible spagetti-code. So I decided to use while and brakes to avoid too long if-else to get to the SorryMessage: 
image

src/main/java/by/custom/utilcalculator/controller/MessageRouter.java
Y-Pazniak: sorry, don't understand you here - need explanation

src/main/java/by/custom/utilcalculator/domain/tree/HelperTree.java
Y-Pazniak: Commands are better than Strings, you explained me why. 
Replaced all Strings to Commands

src/main/java/by/custom/utilcalculator/domain/tree/HelperTree.java
Y-Pazniak: I tried, but it demands URISyntaxException - it doesn't work even if I inherit my own exception from URISyntaxException
image

src/main/java/by/custom/utilcalculator/domain/tree/Node.java
Y-Pazniak: Fixed it. I don't know what happened - but now it works.

src/main/java/by/custom/utilcalculator/exception/TreeReadingException.java
Y-Pazniak: Renamed, thanks

src/main/java/by/custom/utilcalculator/service/UserProgressManager.java
Y-Pazniak: Fixed it according to your advice, thank you

src/main/java/by/custom/utilcalculator/service/UserProgressManager.java
Y-Pazniak: We have an agreement always use {} after if and else, even if it is not necessary.
I removed {} but probably we need to discuss it during our call

src/main/java/by/custom/utilcalculator/service/UserProgressManager.java
Y-Pazniak: It was impossible to use switch-case because I used Strings - it is forbidden to use String in switch-case. 
Using string here was a mistake, I replaced it by switch-case.

Infinity loop was needed here because it made code more beautiful - I hadn't use long 
if(){
}else{
if() {
}else {
if()}}

src/main/resources/words_ru.properties
Y-Pazniak: Fixed it, thank you

src/main/java/by/custom/utilcalculator/controller/Bot.java
MaximPilipeiko: What about System.getProperty.
https://stackoverflow.com/questions/7054972/java-system-properties-and-environment-variables

src/main/java/by/custom/utilcalculator/controller/MessageRouter.java
```diff
@@ -35,36 +35,47 @@ public SendMessage route(final Update update) {
...
-                    answer = userProgressManager.processOwnerType(usersMessage, chatID);
-            case Command.LESS_3_YEARS_AGE, Command.BETWEEN_3_AND_7_YEARS_AGE ->
-                    answer = userProgressManager.processCarAge(usersMessage, chatID);
-            case Command.GASOLINE_TYPE_ENGINE, Command.ELECTRIC_TYPE_ENGINE ->
-                    answer = userProgressManager.processEngineType(usersMessage, chatID);
-            case Command.VOLUME_LESS_1000_CM, Command.VOLUME_BETWEEN_1000_2000_CM, Command.VOLUME_BETWEEN_2000_3000_CM,
-                    Command.VOLUME_BETWEEN_3000_3500_CM, Command.VOLUME_MORE_3500_CM ->
-                    answer = userProgressManager.processEngineVolume(usersMessage, chatID);
+
+            case EAES, OTHER_COUNTRIES -> answer = userProgressManager.processCarOrigin(requestingCommand, chatID);
```
MaximPilipeiko, line 61: What about `break`? Is it not needed?

src/main/java/by/custom/utilcalculator/domain/constants/Command.java
```diff
@@ -1,33 +1,71 @@
...
+    Command (final String command) {
+        this.command = command;
+    }
+
+    public String getCommand(){
+        return command;
+    }
+
+    public static Command getCommandByKey(final String key){
+        for (Command command1 : Command.values()) {
```
MaximPilipeiko: `command1` is not enterprise like naming

src/main/java/by/custom/utilcalculator/domain/tree/HelperTree.java
MaximPilipeiko: Throw, not catch ;)
YOu have to catch URISyntaxException and throw TreeReadingException

MaximPilipeiko (changes requested):

src/main/java/by/custom/utilcalculator/controller/MessageRouter.java
Y-Pazniak: Thanks for asking! I have googled it and discovered this.

1. If we are using traditional switch-case block like 
switch (smth) {
case 1: 
           dosmth(1;) 
           break;
case 2: 
           dosmth(2); 
           break;
case 3:
   dosmth(3);
   break;
}
Using break; is NECESSARY to avoid doing useless code.

2. If we use arrows like 
switch(smth) {
case 1 -> dosmth(1);
case 2 -> dosmth(2);
case 1 -> dosmth(3);
}
Using break; is NOT necessary. 

I didn't know about it.

src/main/java/by/custom/utilcalculator/controller/Bot.java
Y-Pazniak: It's amazing! Everything works fine, thanks a lot!
If you want to set your own Properties in Idea: three points (icon in the top right corner) - edit - modify options - add VM options - set pair key-value (example: -Dtest=test).

Also I have just read about the difference between properties and env variables and now I understand better why System.properties are better (they work only for JVM and do not affect on OS)

src/main/java/by/custom/utilcalculator/domain/constants/Command.java
Y-Pazniak: Fixed

src/main/java/by/custom/utilcalculator/domain/tree/HelperTree.java
Y-Pazniak: Fixed, thank you

src/main/java/by/custom/utilcalculator/controller/Bot.java
```diff
@@ -24,6 +24,7 @@ public static Bot getBot() {
     private Bot() {
         //initDB();
         messageRouter = MessageRouter.getInstance();
+        CommandTree.getInstance();
```
MaximPilipeiko, line 27: Please, add a comment on why you call getInstance without initializing any variable like in the line above.

src/main/java/by/custom/utilcalculator/controller/MessageRouter.java
```diff
@@ -35,36 +35,49 @@ public SendMessage route(final Update update) {
...
-                    answer = userProgressManager.processEngineVolume(usersMessage, chatID);
+
+            case M1 -> answer = userProgressManager.processTransportType(requestingCommand, chatID);
+
+            case EAES, OTHER_COUNTRIES -> answer = userProgressManager.processCarOrigin(requestingCommand, chatID);
+
+
+            case PHYSICAL_PERSON, JURIDICAL_PERSON ->
+                    answer = userProgressManager.processOwnerType(requestingCommand, chatID);
+
```
MaximPilipeiko, line 66: too many empty lines to my taste. It's not call to action

MaximPilipeiko (approved):

src/main/java/by/custom/utilcalculator/controller/Bot.java
Y-Pazniak: Comment was added

src/main/java/by/custom/utilcalculator/controller/MessageRouter.java
Y-Pazniak: Agree, the code had been cleaned

PR #21, opened 2024-07-15
New branch "except M1" 

Y-Pazniak:
M1 branch works with new branches - need to fill another branches.


src/main/java/by/custom/utilcalculator/service/UserProgressManager.java
```diff
@@ -134,26 +169,61 @@ public String processEngineType(final Command requestingCommand, final String ch
...
-            case Command.VOLUME_BETWEEN_3000_3500_CM ->
-                    userProgress.setVolumeOfEngine(VolumeOfEngine.BETWEEN_3000_AND_3500);
-            case VOLUME_MORE_3500_CM -> userProgress.setVolumeOfEngine(VolumeOfEngine.MORE_3500);
+            case TRUCK_UNITS_6_CLASS -> userProgress.setTruckUnitType(TruckUnit.TRUCK_UNITS_6_CLASS);
+            case TRUCK_UNITS_EXCEPT_6_CLASS -> userProgress.setTruckUnitType(TruckUnit.TRUCK_UNITS_EXCEPT_6_CLASS);
         }
         userProgressStorage.save(userProgress);
-        message = messagesCreator.getSummaryAnswer(userProgress);
-        return message;
+        System.out.println("truck units");
```
MaximPilipeiko: Is it needed?

src/main/resources/tree.json
```diff
@@ -78,9 +73,101 @@
...
+                }
+              ]
+            }
+          ]
+        },
+        {
+          "key": "TRUCK_UNITS",
+          "children": [
+            {
+              "key": "TRUCK_UNITS_EXCEPT_6_CLASS",
```
MaximPilipeiko: If I understand correctly what you would like to achieve, It's better to use for this cases `_OTHER`
```suggestion
              "key": "TRUCK_UNITS_OTHER",
```

src/main/resources/tree.json
```diff
@@ -78,9 +73,101 @@
     },
     {
       "key": "EXCEPT_M1",
```
MaximPilipeiko:
```suggestion
      "key": "OTHER",
```

MaximPilipeiko: Let's disscuss it on a call

src/main/java/by/custom/utilcalculator/service/UserProgressManager.java
```diff
@@ -134,26 +169,90 @@ public String processEngineType(final Command requestingCommand, final String ch
...
         }
         userProgressStorage.save(userProgress);
-        message = messagesCreator.getSummaryAnswer(userProgress);
-        return message;
+        return messagesCreator.getSummaryAnswer(userProgress);
+    }
+
+    public String processTruckUnitWeight(final Command requestingCommand, final String chatID) throws UtilsborException {
+        final UserProgress userProgress = userProgressStorage.get(chatID);
+        if (!UserProgressValidator.validateCommand(Command.TRUCK_UNIT_WEIGHT, userProgress)) {
```
MaximPilipeiko: Is command really should be hardocoded?
```suggestion
        if (!UserProgressValidator.validateCommand(requestingCommand, userProgress)) {
```

src/main/java/by/custom/utilcalculator/service/UserProgressManager.java
```diff
@@ -134,26 +169,90 @@ public String processEngineType(final Command requestingCommand, final String ch
...
+            case TRUCK_UNITS_20_50_TONS -> userProgress.setTruckUnitWeight(TruckUnitWeight.FROM_20_TILL_50_TONS);
+        }
+
+        userProgressStorage.save(userProgress);
+        return messagesCreator.getSummaryAnswer(userProgress);
+    }
+
+    public String processTrailersO4Type(final Command requestingCommand, final String chatID) throws UtilsborException {
+        final UserProgress userProgress = userProgressStorage.get(chatID);
+        if (!UserProgressValidator.validateCommand(Command.TRAILERS_O4_TYPE, userProgress)) {
```
MaximPilipeiko: Is command really should be hardocoded?
```suggestion
        if (!UserProgressValidator.validateCommand(requestingCommand, userProgress)) {
```

src/main/java/by/custom/utilcalculator/service/CalculatorRouter.java
MaximPilipeiko, line 1: Let's disscuss it

MaximPilipeiko (changes requested):
Looks good. Just need not discuss to understand some things

src/main/java/by/custom/utilcalculator/service/UserProgressManager.java
Y-Pazniak: No, we don't need sout here. I don't have it in my last version - probably it is an old commit or smth - I don't have any System commands in my code
image

src/main/resources/tree.json
Y-Pazniak: OK, let's discuss. In my opinion "other" is too wide - we can use it everywhere, but "except m1" is rather narrow and when I see "except m1" I clearly understand, that it is any type of vehicle except passenger auto. Probably, we can use "m1_vehicles" and "other_vehicles"?

src/main/resources/tree.json
Y-Pazniak: In this case yes, we can rename - we have only 2 types of trucks: 6 class and not-6-class.
Renamed

src/main/java/by/custom/utilcalculator/service/UserProgressManager.java
Y-Pazniak: In this case yes, thank you for noticing it. 
Sometimes I have a node with different answer. Example: 
image

But i don't care about user answer - I know that despite user answer he has only one possible next step. 
Example.
image
 
Whatever user choose, next step - age auto.
For this cases I decided to unite commands in one common command to avoid useless extra nodes in tree. 
image
This is possibilities for user commands. 

image
This is how it looks in the tree. 

And I mustn't check the particular command, I have to check only does this command belongs to correct group of commands or not and I need to hardcode this command.

src/main/resources/tree.json
MaximPilipeiko: Please, add description

src/main/java/by/custom/utilcalculator/controller/Bot.java
```diff
@@ -22,8 +22,9 @@ public static Bot getBot() {
 
     //private constructor to avoid wrong bot's creation
     private Bot() {
-        //initDB();
+        //need MessageRouter object to work with it after - to proceed user's messages
         messageRouter = MessageRouter.getInstance();
+        //don't need commandTree object in future - only to load tree.json file to build correct branches of calculator
         CommandTree.getInstance();
```
MaximPilipeiko, line 27:
```suggestion
        CommandTree.init();
```

src/main/java/by/custom/utilcalculator/domain/constants/Command.java
```diff
@@ -4,67 +4,95 @@ public enum Command {
...
+    M1_GASOLINE_ENGINE_VOLUME("volume"),
+    N1_N3_WEIGHT("n1_n3_weight"),
+    M2_M3_GASOLINE_ENGINE_VOLUME("m2_m3_volume"),
+    TRUCK_UNIT_WEIGHT("truck_unit_weight"),
+    TRAILERS_O4_TYPE("trailers_o4_weight");
 
     private final String command;
 
-    Command (final String command) {
+    Command(final String command) {
```
MaximPilipeiko:
```suggestion
    Command(final String command, final @Nullable Command reference) {
```

src/main/java/by/custom/utilcalculator/domain/constants/steps/ExceptM1TransportType.java
```diff
@@ -0,0 +1,5 @@
+package by.custom.utilcalculator.domain.constants.steps;
+
+public enum ExceptM1TransportType implements StepsIndicator {
```
MaximPilipeiko: Please, find and change ExceptM1 to smth like BusesAndTrucks
```suggestion
public enum BusesAndTrucksTransportType implements StepsIndicator {
```

src/main/java/by/custom/utilcalculator/domain/constants/steps/GeneralTransportType.java
```diff
@@ -0,0 +1,5 @@
+package by.custom.utilcalculator.domain.constants.steps;
+
+public enum GeneralTransportType implements StepsIndicator{
+    M1, EXCEPT_M1, SELF_PROPELLED_VEHICLES
```
MaximPilipeiko:
```suggestion
    M1, BUSSES_AND_TRUCKS, SELF_PROPELLED_VEHICLES
```

src/main/java/by/custom/utilcalculator/domain/constants/Command.java
```diff
@@ -1,70 +1,117 @@
...
+    //trailers O4 types
+    TRAILERS_04_TYPE("/trailers_o4_type", TRAILERS_O4_TYPE, Step.CAR_AGE),
+    HALF_TRAILERS_04_TYPE("/semi_trailers_o4_type", TRAILERS_O4_TYPE, Step.CAR_AGE);
 
     private final String command;
+    private final Command family;
+    private final Step nextStep;
 
-    Command (final String command) {
+    Command(final String command, final Command family, final Step step) {
```
MaximPilipeiko: Please, rememeber me what we decided to store in `step`. If it's the next step it might be a problem. We may not know which language will be used at this level. The next step is defined in `tree.json`

src/main/java/by/custom/utilcalculator/service/BundleResourcesServant.java
```diff
@@ -26,7 +19,12 @@ private static class BundleResourcesCreatorHolder {
     }
 
     public String getString(final String stringToGetFromBundle) {
-        return bundle.getString(stringToGetFromBundle);
+        if (bundle.containsKey(stringToGetFromBundle)) {
+            return bundle.getString(stringToGetFromBundle);
+        } else {
+            System.out.println("no such string in bundle, check the path");
```
MaximPilipeiko, line 25: Which path?
```suggestion
            System.out.printf("no such string in bundle, check the path `%s`\n", stringToGetFromBundle);
```

src/main/java/by/custom/utilcalculator/service/BundleResourcesServant.java
```diff
@@ -26,7 +19,12 @@ private static class BundleResourcesCreatorHolder {
     }
 
     public String getString(final String stringToGetFromBundle) {
```
MaximPilipeiko, line 21: I think this name is more attractive and informative enough.
```suggestion
    public String getString(final String localizationKey) {
```

src/main/java/by/custom/utilcalculator/service/UserProgressManager.java
```diff
@@ -32,25 +32,74 @@ public void createNewUserProgress(final String chatID) throws UtilsborException
...
+
+    public String processN1_N3TransportWeight(final Command requestingCommand, final String chatID) throws UtilsborException {
+        final UserProgress userProgress = userProgressStorage.get(chatID);
+
+        if (!UserProgressValidator.validateCommand(requestingCommand, userProgress)) {
+            throw new InvalidOrderCommandException(chatID, requestingCommand);
+        }
+        userProgress.setNextStep(requestingCommand.getNextStep());
+        switch (requestingCommand) {
+            case LESS_2_TONS -> userProgress.setTransportWeightN1N2N3(N1N3TransportWeight.LESS_2_TONS);
```
MaximPilipeiko: You have already create association between Command and StepIndicator in `fillFieldsToCommandsMap`. Please, use this map instead of duplication. It will help to avoid potential inconsistency if smth will be missed.

src/main/java/by/custom/utilcalculator/domain/UserProgress.java
```diff
@@ -8,50 +8,118 @@
...
+    private ExceptM1TransportType exceptM1TransportType = null;
+    private N1N3TransportWeight transportWeightN1N2N3 = null;
+    private M2M3EngineType engineTypeM2M3 = null;
+    private M2EngineVolume m2EngineVolume = null;
+    private TruckUnitClass truckUnitClass = null;
+    private TruckUnitWeight truckUnitWeight = null;
+    private TrailerO4Type trailersO4Type = null;
     private final String chatID;
-    private Step currentQuestion;
+    private Step nextStep;
```
MaximPilipeiko: Not sure that it's good decision store nextStep instead of current

src/main/java/by/custom/utilcalculator/domain/UserProgress.java
```diff
@@ -8,50 +8,118 @@
...
+
+    public GeneralTransportType getGeneralTransportType() {
+        return generalTransportType;
     }
 
     public CountryOrigin getCountryOrigin() {
         return countryOrigin;
     }
 
     public Command[] getUserPath() {
```
MaximPilipeiko: Already forgot how we use the user path. Please, remember me to discuss it on our call. Now it looks strange.

src/main/resources/tree.json
```diff
@@ -1,28 +1,33 @@
 {
   "key": "START",
```
MaximPilipeiko, line 2: Could you please, draw a visualization of this tree? Even physical on paper will be enough.

src/main/java/by/custom/utilcalculator/domain/UserProgress.java
MaximPilipeiko: Too many steps. Possible we have to fill it after each user command and not calculate it. We will discuss it.

src/main/java/by/custom/utilcalculator/domain/UserProgress.java
```diff
@@ -71,75 +137,47 @@ public CarAge getCarAge() {
...
+    public TruckUnitClass getTruckUnitType() {
+        return this.truckUnitClass;
+    }
+
+    public String getChatID() {
+        return chatID;
     }
 
-    private void cleanStepsAfterCurrent(final int stepCleaner) {
+    private void cleanStepsAfterCurrentM1Branch(final int stepCleaner) {
```
MaximPilipeiko: It's strange that we have two different methods `cleanStepsAfterCurrentM1Branch` and `cleanStepsAfterCurrentExceptM1Branch` instead of one `cleanStepsAfterCurrent `

MaximPilipeiko (changes requested):

src/main/java/by/custom/utilcalculator/controller/Bot.java
Y-Pazniak: Fixed.

src/main/java/by/custom/utilcalculator/domain/constants/Command.java
Y-Pazniak: Done.

src/main/java/by/custom/utilcalculator/domain/constants/steps/ExceptM1TransportType.java
Y-Pazniak: Done.

src/main/java/by/custom/utilcalculator/domain/constants/steps/GeneralTransportType.java
Y-Pazniak: Fixed.

src/main/java/by/custom/utilcalculator/domain/constants/Command.java
Y-Pazniak: We decided to store there current step. My personal decision was to store there next step. Possibly, I was wrong.

src/main/java/by/custom/utilcalculator/service/BundleResourcesServant.java
Y-Pazniak: Fixed.

src/main/java/by/custom/utilcalculator/service/BundleResourcesServant.java
Y-Pazniak: Renamed.

src/main/java/by/custom/utilcalculator/service/UserProgressManager.java
Y-Pazniak: I like your idea very much. Unfortunately, it seems I have a problem.
We have userProgress object with fields for price calculation. We need to fill fields with 100% accuracy. We need to know which type of vehicle in particular to calculate a correct price. 
But during filling fillFieldsToCommandsMap sometimes we use families instead of particular field. 
For example. For a price calculation I have to know the volume of an engine. But for the fillFieldsToCommandsMap is enough to know only that volume field was filled - no difference, which in particular. 
I see no solutions right now, but we can discuss it and find the best way.

src/main/java/by/custom/utilcalculator/domain/UserProgress.java
Y-Pazniak: Highly likely you are right. It was completely my decision, I disobeyed your order to save here current step

src/main/java/by/custom/utilcalculator/domain/UserProgress.java
Y-Pazniak: I use getUserPath().
1. For getting the current user's node. Class CommandTree - method getNode(final UserProgress userProgress) returns current user's node. 
2. For user path string building (a line about previous user's steps).
3. For debugging to check this array filling - what commands there are.

src/main/java/by/custom/utilcalculator/domain/UserProgress.java
Y-Pazniak: Now we have three of them :) 
I can try to rewrite it into one and to check what will happen.

src/main/resources/tree.json
Y-Pazniak: [chatbot-scheme.pdf](https://github.com/user-attachments/files/16522953/chatbot-scheme.pdf)

src/main/java/by/custom/utilcalculator/domain/UserProgress.java
```diff
@@ -5,155 +5,200 @@
...
-        final int stepID = 0;
-        cleanStepsAfterCurrent(stepID);
-        currentQuestion = Step.TRANSPORT_TYPE;
-        this.transportType = transportType;
+    public TrailerO4Type getTrailerO4Type() {
+        return trailersO4Type;
+    }
+
+    public void setTrailerO4Type(final TrailerO4Type trailersO4Type) {
+        cleanStepsAfterCurrentM1Branch(3);
```
MaximPilipeiko: You have to calculate the step number. I expect to see something like this.
```suggestion
        var newStep = TreeMediator.getNextStep(this);
        cleanStepsAfterCurrent(newStep.number);
        this.trailersO4Type = trailersO4Type;
        updateUserPath(trailersO4Type);
```

src/main/java/by/custom/utilcalculator/domain/UserProgress.java
```diff
@@ -5,155 +5,200 @@
...
-    private Step currentQuestion;
+    private final List<Command> userPath;
 
     public UserProgress(final String chatID) {
         this.chatID = chatID;
+        userPath = new ArrayList<>();
+    }
+
+    private void addUserStatusToPath(final StepsIndicator stepsIndicator) {
+        final Map<StepsIndicator, Command> fieldsToCommands = CommandTree.init().getFieldsToCommands();
```
MaximPilipeiko: These 3 lines should be enough I believe.
```suggestion
        final Map<StepsIndicator, Command> fieldsToCommands = CommandTree.init().getFieldsToCommands();
        Command command = fieldsToCommands.get(stepsIndicator);
        userPath.add( command);
```

src/main/java/by/custom/utilcalculator/domain/tree/CommandTree.java
```diff
@@ -19,18 +21,24 @@ private CommandTree() throws UtilsborCommandTreeReadingException {
         treeRoot = HelperTree.buildTree();
     }
 
-    public static CommandTree getInstance() {
+    public static CommandTree init() {
```
MaximPilipeiko:
```suggestion
    public static CommandTree getInstance() {
```

src/main/java/by/custom/utilcalculator/domain/TreeMediator.java
```diff
@@ -0,0 +1,16 @@
+package by.custom.utilcalculator.domain;
+
+import by.custom.utilcalculator.domain.constants.Command;
+import by.custom.utilcalculator.domain.constants.steps.Step;
+import by.custom.utilcalculator.domain.tree.CommandTree;
+
+public class TreeMediator {
+    public static boolean validateCommand(final Command requestingCommand, final UserProgress userProgress) {
```
MaximPilipeiko: I think this method can be moved to `CommandTree`. Additional class `TreeMediator` is redundant.

src/main/java/by/custom/utilcalculator/domain/TreeMediator.java
```diff
@@ -0,0 +1,16 @@
...
+        Command command = requestingCommand.getFamily() == null ? requestingCommand : requestingCommand.getFamily();
+        return CommandTree.init().validateCommandFromNode(command, userProgress);
+    }
+
+    public static Step getNextStep (final UserProgress userProgress){
```
MaximPilipeiko: I think this method can be moved to `CommandTree`. Additional class `TreeMediator` is redundant.

src/main/java/by/custom/utilcalculator/domain/constants/Command.java
```diff
@@ -1,70 +1,111 @@
 package by.custom.utilcalculator.domain.constants;
 
+import javax.annotation.Nullable;
+
 public enum Command {
+    //command families - not commands actually, just markers
+    AGE("age", null),
+    ENGINE_VOLUME("volume", null),
+    N1_N3_WEIGHT("n1_n3_weight", null),
```
MaximPilipeiko: I think you can merge   N1_N3_WEIGHT and WEIGHT into one WEIGHT. But for now, lets leave as it's.

src/main/java/by/custom/utilcalculator/domain/tree/CommandTree.java
```diff
@@ -19,18 +21,24 @@ private CommandTree() throws UtilsborCommandTreeReadingException {
...
         return fieldsToCommands;
     }
 
-    public boolean validateCommand(final Command requestingCommand, final UserProgress userProgress) {
+    public boolean validateCommandFromNode(final Command requestingCommand, final UserProgress userProgress) {
         return isRequestingCommandAcceptable(requestingCommand, getNode(userProgress));
     }
 
+    public Step getNextStepFromNode(final UserProgress userProgress) {
+        System.out.println("current node: " + getNode(userProgress).getKey());
```
MaximPilipeiko: Please, rid `println`

src/main/java/by/custom/utilcalculator/domain/tree/Node.java
```diff
@@ -10,12 +11,17 @@ public class Node {
     private Node parent;
     private final List<Node> children;
     private final Command key;
+    private final Step nextStep;
```
MaximPilipeiko, line 15: You can store step number. Step number is a level of the node in tree.
```suggestion
private Int number;
```

src/main/java/by/custom/utilcalculator/service/CalculatorExceptM1.java
```diff
@@ -0,0 +1,155 @@
...
+
+    private static class CalculatorExceptM1Holder {
+        private static final CalculatorExceptM1 CALCULATOR_INSTANCE = new CalculatorExceptM1();
+    }
+
+    public static CalculatorExceptM1 getInstance() {
+        return CalculatorExceptM1Holder.CALCULATOR_INSTANCE;
+    }
+
+    public String calculate(final UserProgress userProgress) {
```
MaximPilipeiko, line 19: For me it looks like you can store the final price in the node as well.

MaximPilipeiko (changes requested):

src/main/java/by/custom/utilcalculator/domain/constants/Command.java
Y-Pazniak: It is better to do it now. It was just my mistake to miss this particular family. Fixed

src/main/java/by/custom/utilcalculator/domain/tree/CommandTree.java
Y-Pazniak: I can rid it, but is it necessary for now? It does great help for me during testing the bot: I can see the current step and the next step in console: if I have an error and nodes don't work in a proper way, I can discover this error from console easily. 

It is not the first time you ask me to delete println - why is it important? Let's discuss.

src/main/java/by/custom/utilcalculator/domain/tree/Node.java
Y-Pazniak: Let's discuss, how we can use int stepNumber

src/main/java/by/custom/utilcalculator/service/CalculatorExceptM1.java
Y-Pazniak: Sounds promising, but this is a magic for me. We should discuss ways how is it possible to implement this feature

src/main/java/by/custom/utilcalculator/domain/TreeMediator.java
Y-Pazniak: I can do it easily. I just tried to do less dependencies between 2 big parts of our program: tree and userProgress I decided to use Mediator to avoid direct connections between Node and UserProgress.

src/main/java/by/custom/utilcalculator/domain/tree/CommandTree.java
Y-Pazniak: Fixed

src/main/java/by/custom/utilcalculator/domain/UserProgress.java
Y-Pazniak: In my head we don't need cleanSteps here future - it should works automatically. I'm going to write to you my solution in telegram and I'll be happy to discuss it with you.

src/main/java/by/custom/utilcalculator/domain/TreeMediator.java
MaximPilipeiko: `getNextStepFromNode` and `validateCommandFromNode` are already in CommandTree. For me, one more layer will not reduce dependency.

src/main/java/by/custom/utilcalculator/domain/UserProgress.java
```diff
@@ -5,155 +5,158 @@
...
+    private void addUserStatusToPath(final StepsIndicator stepsIndicator) {
+        final Map<StepsIndicator, Command> fieldsToCommands = CommandTree.getInstance().getFieldsToCommands();
+        Command command = fieldsToCommands.get(stepsIndicator);
+        Class<? extends StepsIndicator> stepsIndicatorClass = stepsIndicator.getClass();
+        boolean needAddCommand = true;
+        int lastNumber = 0;
+
+        for (int i = 0; i < userPath.size(); i++) {
+            Command localCommand = userPath.get(i);
+            for (Map.Entry<StepsIndicator, Command> entry : fieldsToCommands.entrySet()) {
```
MaximPilipeiko, line 40: Looks like `Map<Command, StepsIndicator> commandsToFields` will be useful

src/main/java/by/custom/utilcalculator/domain/UserProgress.java
```diff
@@ -5,155 +5,158 @@
...
+                if (entry.getValue().equals(localCommand)) {
+                    if (entry.getKey().getClass().equals(stepsIndicatorClass)) {
+                        needAddCommand = false;
+                        userPath.set(i, command);
+                        lastNumber = i;
+                        break;
```
MaximPilipeiko: Please, check. It will be broken only from the first `for` loop but I think you would like to finish both loops.

src/main/java/by/custom/utilcalculator/domain/UserProgress.java
```diff
@@ -5,155 +5,158 @@
...
+                    }
+                }
+            }
+        }
+
+        if (needAddCommand) {
+            userPath.add(command);
+        } else {
+            for (int i = lastNumber + 1; i < userPath.size(); i++) { //this code removes all cells in list after requested by user
```
MaximPilipeiko: I suggest to remove all elements from list instead of set null.
https://stackoverflow.com/questions/22802232/remove-all-elements-from-a-list-after-a-particular-index

src/main/java/by/custom/utilcalculator/domain/tree/CommandTree.java
MaximPilipeiko: Let's discuss it on call.

MaximPilipeiko (changes requested):
Almost done

src/main/java/by/custom/utilcalculator/domain/UserProgress.java
Y-Pazniak: _one more layer will not reduce dependency._
Probably you are right. Fixed it. 

_Looks like Map<Command, StepsIndicator> commandsToFields will be useful_
I've added a new map, but probably I've done smth wrong - I feel no big difference between before and after adding the map

src/main/java/by/custom/utilcalculator/domain/UserProgress.java
Y-Pazniak: Yes, great notice. Fixed it, thank you.

src/main/java/by/custom/utilcalculator/domain/UserProgress.java
Y-Pazniak: That's a brilliant idea, much better than nulls. Implemented it. Thanks a lot.

src/main/java/by/custom/utilcalculator/domain/TreeMediator.java
Y-Pazniak: Done

src/main/java/by/custom/utilcalculator/domain/TreeMediator.java
Y-Pazniak: Now I see, done

src/main/java/by/custom/utilcalculator/domain/UserProgress.java
Y-Pazniak: Sorry, don't understand your proposition. Probably, I should explain what's going on in this code.

src/main/java/by/custom/utilcalculator/controller/MessageRouter.java
```diff
@@ -44,7 +44,11 @@ private SendMessage getCheckInputMessageAndGetAnswer(final Message message) {
             e.printStackTrace();
             return new SendMessage(chatID, getExceptionText(e));
         }
-        return new SendMessage(chatID, answer);
+        SendMessage sendMessage = new SendMessage();
```
MaximPilipeiko: Using a constructor rather than setters to create an object is better. Please, read https://www.reddit.com/r/javahelp/comments/geg5ty/constructors_vs_getters_setters/. And we will discuss this question during the call.
return new SendMessage(
  chatID, answer, "HTML"
);

src/main/java/by/custom/utilcalculator/domain/UserProgress.java
```diff
@@ -5,155 +5,158 @@
...
+    private void addUserStatusToPath(final StepsIndicator stepsIndicator) {
+        final Map<StepsIndicator, Command> fieldsToCommands = CommandTree.getInstance().getFieldsToCommands();
+        Command command = fieldsToCommands.get(stepsIndicator);
+        Class<? extends StepsIndicator> stepsIndicatorClass = stepsIndicator.getClass();
+        boolean needAddCommand = true;
+        int lastNumber = 0;
+
+        for (int i = 0; i < userPath.size(); i++) {
+            Command localCommand = userPath.get(i);
+            for (Map.Entry<Command, StepsIndicator> entry : CommandTree.getInstance().getCommandsToFields().entrySet()) {
```
MaximPilipeiko: What about using `fieldsToCommands`?

src/main/java/by/custom/utilcalculator/domain/tree/HelperTree.java
```diff
@@ -75,4 +212,12 @@ private static void fillParents(final Node node) {
             }
         }
     }
+
+    public static Map<Command, StepsIndicator> fillCommandsToFields(Map<StepsIndicator, Command> fieldsToCommands) {
+        Map<Command, StepsIndicator> commandsToFields = new HashMap<>();
```
MaximPilipeiko: Looks like it should be `Map<Command, List<StepsIndicator>>`. Please, look at `Command.ENGINE_VOLUME`.

src/main/java/by/custom/utilcalculator/service/MessagesCreator.java
```diff
@@ -22,22 +35,34 @@ public static MessagesCreator getInstance() {
     public String buildNextStepQuestion(final UserProgress userProgress) {
```
MaximPilipeiko, line 35: Maybe it's also possible to store it in tree

src/main/java/by/custom/utilcalculator/service/UserProgressManager.java
```diff
@@ -26,31 +26,99 @@ public static UserProgressManager getInstance() {
...
+
+    public String processN1_N3TransportWeight(final Command requestingCommand, final String chatID) throws UtilsborException {
+        final UserProgress userProgress = userProgressStorage.get(chatID);
+
+        if (!CommandTree.validateCommand(requestingCommand, userProgress)) {
+            throw new InvalidOrderCommandException(chatID, requestingCommand);
+        }
+
+        switch (requestingCommand) {
+            case LESS_2_TONS -> userProgress.setWeight(Weight.LESS_2_TONS);
```
MaximPilipeiko, line 101: You already have association between command and StepIndecator. I think you can use instead duplicate this in switch-case

MaximPilipeiko (changes requested):

src/main/java/by/custom/utilcalculator/controller/MessageRouter.java
Y-Pazniak: Done. Impossible to read an article for me, let's discuss together

src/main/java/by/custom/utilcalculator/domain/UserProgress.java
Y-Pazniak: I like using fieldsToCommands more.

src/main/java/by/custom/utilcalculator/service/MessagesCreator.java
Y-Pazniak: Can we call somehow a method from the a tree? If it is possible, it is better to replace.

src/main/java/by/custom/utilcalculator/service/UserProgressManager.java
Y-Pazniak: 🤔 it looks very promising, I'll try it

src/main/java/by/custom/utilcalculator/domain/tree/HelperTree.java
Y-Pazniak: I see - we did a mistake using the same key for different values. I'll fix it easily.

src/main/java/by/custom/utilcalculator/controller/Bot.java
```diff
@@ -36,7 +36,7 @@ public String getBotUsername() {
     @Override
     public String getBotToken() {
         return
-                System.getProperty("custom_chatbot_token");
+                System.getProperty("custom_chatbot_test_token");
```
MaximPilipeiko, line 39: It looks strange that you add `_test_` to token name.

src/main/java/by/custom/utilcalculator/domain/UserProgress.java
```diff
@@ -5,155 +5,156 @@
...
 
     public UserProgress(final String chatID) {
         this.chatID = chatID;
+        userPath = new ArrayList<>();
     }
 
-    public TransportType getTransportType(){
-        return transportType;
+    private void addUserStatusToPath(final StepsIndicator stepsIndicator) {
+        final Map<StepsIndicator, Command> fieldsToCommands = CommandTree.getInstance().getFieldsToCommands();
```
MaximPilipeiko, line 32: Now I think it's better to rely on tree to get step number. Let's discuss it on a call.

MaximPilipeiko (approved):

PR #26, opened 2024-10-11
Self propelled vehicles

Y-Pazniak:
The branch contains self-propelled vehicles and other fixes after replacement old code and add new functions of filling and loading tree from java-code


src/main/java/by/custom/utilcalculator/domain/UserProgress.java
```diff
@@ -5,155 +5,156 @@
...
+    private void addUserStatusToPath(final StepsIndicator stepsIndicator) {
+        final Map<StepsIndicator, Command> fieldsToCommands = CommandTree.getInstance().getFieldsToCommands();
+        Command command = fieldsToCommands.get(stepsIndicator);
+        Class<? extends StepsIndicator> stepsIndicatorClass = stepsIndicator.getClass();
+        boolean needAddCommand = true;
+        int lastNumber = 0;
+
+        for (int i = 0; i < userPath.size(); i++) {
+            Command localCommand = userPath.get(i);
+            for (Map.Entry<StepsIndicator, Command> entry : fieldsToCommands.entrySet()) {
```
MaximPilipeiko: As we discussed, passing through a command tree is better than a map.

src/main/java/by/custom/utilcalculator/domain/tree/ModifierTree.java
```diff
@@ -0,0 +1,64 @@
...
+import by.custom.utilcalculator.domain.UserProgress;
+import by.custom.utilcalculator.domain.constants.Command;
+import by.custom.utilcalculator.exception.UtilsborCommandTreeReadingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+
+public class ModifierTree {
```
MaximPilipeiko: It's more correct
```suggestion
public class TreeModifier {
```

src/main/java/by/custom/utilcalculator/domain/tree/ModifierTree.java
```diff
@@ -0,0 +1,64 @@
...
+        if (treeRootJson != null) {
+            fillParents(treeRootJson);
+        }
+        return treeRootJson;
+    }
+
+    public static String getPrice(final UserProgress userProgress) throws UtilsborCommandTreeReadingException {
+        Node root;
+        try {
+            root = mapper.readValue(getTreeInputStream(), Node.class);
```
MaximPilipeiko: I think it's redundant to build a tree here. I think it's enough to build a tree once on application starts up.

src/main/java/by/custom/utilcalculator/domain/tree/ModifierTree.java
```diff
@@ -0,0 +1,64 @@
...
+        }
+
+        Node localNode = root;
+        if (root != null) { //searching for the proper node according to user's path
+            for (Command command : userProgress.getUserPath()) {
+                for (Node node : localNode.getChildren()) {
+                    if (node.getKey().equals(command)) {
+                        localNode = node;
+                        if (node.getPrice() != null) {
+                            return node.getPrice();
```
MaximPilipeiko: looks like you may return price even to not pass all user path.

MaximPilipeiko (changes requested):

src/main/java/by/custom/utilcalculator/domain/UserProgress.java
Y-Pazniak: I did this code better, but I didn't manage to use the Tree here. The main reason - i didn't find the way how to use the tree to move back for several steps and to clean List<Command> userPath. 
Example. User path is: A.class -> B.class -> D.class -> E.class 
User wants to get back to the B.class. 
I need to fall back to B.class and clean all steps after it. I need to start a foreach and find an index of B.class object and then to clean this List userPath.sublist(b.index, userPath.size()).clean(); 
We can not use only the Tree for this purposes - we need foreach. 
Probably I did something wrong but I still don't know how to use the Tree for falling back user's path. We can discuss it.

src/main/java/by/custom/utilcalculator/domain/tree/ModifierTree.java
Y-Pazniak: Renamed. Should we rename also CommandTree to TreeCommand?

src/main/java/by/custom/utilcalculator/domain/tree/ModifierTree.java
Y-Pazniak: Done.

src/main/java/by/custom/utilcalculator/domain/tree/ModifierTree.java
Y-Pazniak: Yes, my gross mistake. Fixed it

src/main/java/by/custom/utilcalculator/domain/tree/TreeModifier.java
```diff
@@ -0,0 +1,56 @@
...
+                    if (node.getKey().equals(command)) {
+                        localNode = node;
+                        if (node.getPrice() != null && node.getChildren().isEmpty()) {
+                            return node.getPrice();
+                        }
+                    }
+                }
+            }
+        }
+        return "error during calculation";
```
MaximPilipeiko: For me, it looks like the exception better fits here.

MaximPilipeiko (approved):

PR #28, opened 2024-10-24
Next message from tree (not switch-cases)

Y-Pazniak:
New option - loading from tree - has been added. Without placeholders for now


MaximPilipeiko:
Could you please change base branch to `self-propelled-vehicles`. For ability to see real changes and avoid review the same change twice.

src/main/resources/tree.json
```diff
@@ -148,17 +172,20 @@
                       "key": "VOLUME_BETWEEN_2000_3000_CM",
                       "nextStep": "AGE",
                       "price": null,
+                      "nextMessage": "questions.m1.other.gasolinePower.2000_3000",
                       "children": [
                         {
                           "key": "LESS_3_YEARS_AGE",
                           "nextStep": "FAREWELL",
                           "price": "<b>32 727,55 </b>",
```
MaximPilipeiko:
```suggestion
                          "price": 32727.55
```

src/main/java/by/custom/utilcalculator/domain/tree/Node.java
```diff
@@ -15,15 +15,18 @@ public class Node {
     private List<Node> children;
     private final Command key;
     private final Step nextStep;
+    private final String nextMessage;
     private final String price;
```
MaximPilipeiko:
```suggestion
    private final Double price;
```

MaximPilipeiko:
```suggestion
    private final Price price;
```

src/main/resources/tree.json
```diff
@@ -2,32 +2,38 @@
...
             {
               "key": "PHYSICAL",
               "nextStep": "AGE",
               "price": null,
+              "nextMessage": "questions.m1.eaes.agePhysical",
               "children": [
                 {
                   "key": "LESS_3_YEARS_AGE",
                   "nextStep": "FAREWELL",
                   "price": "<b>544,5</b>",
```
MaximPilipeiko:
```suggestion
                  "price": "544,5 BYN",
```

src/main/java/by/custom/utilcalculator/service/MessagesCreator.java
```diff
@@ -35,43 +33,17 @@ public static MessagesCreator getInstance() {
...
-            case AGE -> {
-                return getAgeAuto();
-            }
-            case FAREWELL -> {
-                return getResultAndFarewell(userProgress);
-            }
+        String nextMessage;
+        try {
+            nextMessage = bundle.getString(userProgress.getNextMessage());
+        } catch (NullPointerException e) {
```
MaximPilipeiko, line 24: What would happen if another exception were thrown? Maybe better to catch more general exception?
```suggestion
        } catch (Exception e) {
```

src/main/java/by/custom/utilcalculator/service/UserProgressManager.java
```diff
@@ -28,6 +28,18 @@ public void createNewUserProgress(final String chatID) throws UtilsborException
         userProgressStorage.save(userProgress);
     }
 
+    public String processStart(final Command requestingCommand, final String chatID) throws UtilsborException {
+        final UserProgress userProgress;
+        userProgress = userProgressStorage.get(chatID);
+
+        if (!CommandTree.validateCommand(requestingCommand, userProgress)) {
```
MaximPilipeiko: It's better if boolean is return type.
```suggestion
        if (!CommandTree.isValidCommand(requestingCommand, userProgress)) {
```

MaximPilipeiko (changes requested):

src/main/resources/tree.json
Y-Pazniak: Done.

src/main/java/by/custom/utilcalculator/domain/tree/Node.java
Y-Pazniak: Done.

Y-Pazniak: Done.

src/main/java/by/custom/utilcalculator/service/MessagesCreator.java
Y-Pazniak: Done, but I would like to ask you about  explanations. 
We are requesting a string from a bundle and we can receive this string (if it exists in bundle) or we can not receive a string (if there is not such string in bundle). In my opinion another exceptions here are redundant - the only possible exception here is the NPE.

src/main/java/by/custom/utilcalculator/service/UserProgressManager.java
Y-Pazniak: Done.

src/main/java/by/custom/utilcalculator/domain/Price.java
```diff
@@ -0,0 +1,26 @@
+package by.custom.utilcalculator.domain;
+
+import by.custom.utilcalculator.domain.constants.CurrencyType;
+
+public class Price {
+    private final double amount;
```
MaximPilipeiko: Think about using `BigDecimal` instead double. It's not a fact that it's really needed.
https://medium.com/@cancerian0684/which-data-type-would-you-choose-for-storing-currency-values-like-trading-price-dd7489e7a439

src/main/java/by/custom/utilcalculator/domain/constants/CurrencyType.java
```diff
@@ -0,0 +1,15 @@
+package by.custom.utilcalculator.domain.constants;
+
+public enum CurrencyType {
```
MaximPilipeiko: Where it's possibly better to avoid words that don't give additional sense.
```suggestion
public enum Currency {
```

src/main/java/by/custom/utilcalculator/domain/tree/CommandTree.java
```diff
@@ -9,14 +10,13 @@
 import java.util.stream.Collectors;
 
 public class CommandTree {
-    private final Map<StepsIndicator, Command> fieldsToCommands;
-    private final Map<Command, List<StepsIndicator>> commandsToFields;
+    //private final Map<StepsIndicator, Command> fieldsToCommands;
```
MaximPilipeiko: Add comments why it's commented

src/main/java/by/custom/utilcalculator/domain/tree/ModifierTree.java
```diff
@@ -46,7 +47,7 @@ public static String getPrice(final UserProgress userProgress) throws UtilsborCo
                 }
             }
         }
-        return "error during calculation";
+        return null;
```
MaximPilipeiko, line 50: good

src/main/java/by/custom/utilcalculator/service/MessagesCreator.java
MaximPilipeiko: If you are sure that only NPE is possible, you can leave NullPointerException of course.

MaximPilipeiko (approved):

src/main/java/by/custom/utilcalculator/domain/tree/CommandTree.java
Y-Pazniak: In this case - we refused using maps, now we use nodes to build assotiations between COMMAND and STEP. Don't need maps anymore. Probably, I saved this maps for safety - if something go wrong in future with big branches, I can return maps easily. But it seems everything works fine - I guess, we can delete this maps.

src/main/java/by/custom/utilcalculator/domain/constants/CurrencyType.java
Y-Pazniak: Done.

src/main/java/by/custom/utilcalculator/domain/Price.java
Y-Pazniak: Thanks for the article. I understood, that for finance we should use BigDecimal for accuracy of calculations (and changed double to BigDecimal :)

